PR #44, opened 2015-04-01
allow the use of rufus scheduler methods [at, in]

timcase:
Rufus scheduler allows the scheduling of jobs to occur one at a specific time such as in the case of an auction that is ending. This pull request adds the specific time methods "at" and "in" as allowable options for sidekiq scheduler.


artem-russkikh:
+1

jimryan:
Just out of curiosity, why not just use Sidekiq's native `#perform_at`/`#perform_in`? https://github.com/mperham/sidekiq/wiki/Scheduled-Jobs

Issue #41, opened 2014-10-06
Multiple schedules are loaded for multiple workers


ygoldshtrakh:
We have an environment where we have multiple worker processes and each of this process loads up the scheudler
ie we have:
scheduler: bundle exec sidekiq -c 5 -q scheduler -q default
worker: bundle exec sidekiq -c 5 -q default

So we end up with duplicate schedules , how can we avoid it?

wcurtis:
+1 

Same issue here, using @jgeggatt's fix for now.

artem-russkikh:
Take a look at https://github.com/itbeaver/sidekiq-scheduler/pull/1

```ruby
# config/initializers/sidekiq_scheduler.rb
require 'sidekiq/scheduler'

puts "Sidekiq.server? is #{Sidekiq.server?.inspect}"
puts "defined?(Rails::Server) is #{defined?(Rails::Server).inspect}"
puts "defined?(Unicorn) is #{defined?(Unicorn).inspect}"

if Rails.env == 'production' && (defined?(Rails::Server) || defined?(Unicorn))
  Sidekiq.schedule = YAML
    .load_file(File.expand_path('../../../config/scheduler.yml', __FILE__))
  Sidekiq::Scheduler.reload_schedule!
else
  Sidekiq::Scheduler.enabled = false
  puts "Sidekiq::Scheduler.enabled is #{Sidekiq::Scheduler.enabled.inspect}"
end
```

I use this gem with Unicorn, him process start sidekiq-scheduler and queue tasks
But sidekiq workers disables it when starting (in initializer)

After writing test I make a Merge Request into this repo

PR #49, opened 2015-10-06
Add startup and enabled options

artem-russkikh:
Help to resolve https://github.com/moove-it/sidekiq-scheduler/issues/41
Needs to rewrite README

### DIFF
--- a/test/lib/sidekiq/scheduler_test.rb
+++ b/test/lib/sidekiq/scheduler_test.rb
@@ -4,11 +4,29 @@ class ManagerTest < Minitest::Test
   describe 'Sidekiq::Scheduler' do
 
     before do
+      Sidekiq::Scheduler.enabled = true
       Sidekiq::Scheduler.dynamic = false
       Sidekiq.redis { |r| r.del(:schedules) }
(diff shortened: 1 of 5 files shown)


lib/sidekiq-scheduler/version.rb
```diff
@@ -1,3 +1,3 @@
 module SidekiqScheduler
-  VERSION = '1.0'
+  VERSION = '1.1'
 end
```
elpic: Could you please remove the version bump, It's already bumped.

test/lib/sidekiq/scheduler_test.rb
```diff
@@ -4,6 +4,8 @@ class ManagerTest < Minitest::Test
   describe 'Sidekiq::Scheduler' do
 
     before do
+      Sidekiq::Scheduler.startup = true
```
elpic: Could you please add more tests around the new options?

elpic:
Please squash all the commits into one.

artem-russkikh:
@elpic I remove 'startup' option, because it is enough 'enabled' option for resolving #41 

'enabled' option is covered by one spec, what kind of test can i write more?

PR #60, opened 2015-11-02
Fixed style and complexity issues

elpic:
Fixed issues found by rubocop.

* Increased the Quality GPA

### DIFF
--- a/lib/sidekiq-scheduler.rb
+++ b/lib/sidekiq-scheduler.rb
@@ -6,12 +6,11 @@ require_relative 'sidekiq-scheduler/manager'
 Sidekiq.configure_server do |config|
 
   config.on(:startup) do
-    enabled = Sidekiq::Scheduler.enabled
     scheduler_options = {
-      :scheduler => config.options.fetch(:scheduler, true),
-      :dynamic => config.options.fetch(:dynamic, false),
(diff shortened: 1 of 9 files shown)


lib/sidekiq-scheduler.rb
```diff
@@ -8,10 +8,10 @@
   config.on(:startup) do
     enabled = Sidekiq::Scheduler.enabled
     scheduler_options = {
-      :scheduler => config.options.fetch(:scheduler, true),
-      :dynamic => config.options.fetch(:dynamic, false),
-      :enabled => enabled.nil? ? true : enabled,
-      :schedule => config.options.fetch(:schedule, nil)
+      scheduler: config.options.fetch(:scheduler, true),
+      dynamic:   config.options.fetch(:dynamic, false),
+      enabled:   enabled.nil? ? true : enabled,
```
snmgian: enabled || false

elpic: This is a option that was added recently, and no all configuration have this option, that is way if the option is not present we assume that the scheduler is enabled.

snmgian: Other config options are read though config option, shouldn't be better to follow the same approach?

elpic: Yes that is better. 

Done

Issue #65, opened 2015-11-21
Support Sidekiq >= 4


emerleite:
https://github.com/mperham/sidekiq/blob/master/4.0-Upgrade.md

hesham:
Hey, I added pull request #66 for this.

PR #66, opened 2015-11-22
Upgrade to sidekiq 4.x

hesham:
Also removed celluloid dependency since it's not longer required in sidekiq 4.0.

All tests pass. Please let me know if I missed anything.

### DIFF
--- a/lib/sidekiq-scheduler/manager.rb
+++ b/lib/sidekiq-scheduler/manager.rb
@@ -1,4 +1,3 @@
-require 'celluloid'
 require 'redis'
 
 require 'sidekiq/util'
@@ -14,7 +13,6 @@ module SidekiqScheduler
   #
   class Manager
     include Sidekiq::Util
-    include Celluloid
 
     def initialize(options={})
       @enabled = options[:scheduler]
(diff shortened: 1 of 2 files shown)


reneklacan:
:+1:

elpic:
@hesham 

Would be possible to use the >= 3 for the sidekiq version?

Thanks

hesham:
@stephankaag @elpic Good point. Honestly, I wasn't sure if removing Celluloid dependency will render the gem incompatible with sidekiq 3.x. That said, all tests seem to pass with >= 3. So I'll make that change.

elpic:
I think i was an old code, the celluloid dependency is not necessary on neither of those versions.

hesham:
@elpic Awesome. I made the change.

elpic:
@hesham 

Could you squash the commits? 

Thanks.

hesham:
@elpic yep

elpic:
closes #65

emerleite:
Any plan to a 1.2.5 with this update?

elpic:
I want to update this with the ability to dynamically update the schedule and release the 2.0 version. Probably later this week.

stephankaag:
:+1:

elpic:
@emerleite sorry, I forgot to mention that the version 2.0.0 was released. That version is compatible with the new version of sidekiq and include the dynamic update of schedules.

Issue #34, opened 2014-05-20
Schedules aren't pulled from Redis when changed


stephankaag:
With `dynamic` option set to true I would expect my Sidekiq proces to regularly pull changed schedules from Redis.

`Sidekiq::Scheduler.update_schedule` looks like it is supposed to do so but it's never called (except for a unit test).

lucasgomide:
:+1: works for me, but i`m waiting for merge

Issue #47, opened 2015-05-25
Sidekiq-scheduler reload schedule 


lechen26:
Hi

i'm using Sidekiq 3.1.4 with Sidekiq-scheduler 1.2 with Ruby 2.3 and i'm having issues Reloading my schedules. 
my basic sidekiq.rb looks like:

    #./sidekiq.rb
    Sidekiq.configure_server do |config|
        config.redis = {  url: 'REDIS_URL', namespace: "sidekiq" , :size => 5}
    end
    Dir["./myworker.rb"].each { |file| require file }
    Sidekiq::Scheduler.dynamic = true
    Sidekiq.schedule = YAML.load_file(File.expand_path('../schedule.yml',__FILE__))

    #./sidekiq.yml
    :concurrency: 25
    :pidfile: ./sidekiq.pid
    :logfile: ./sidekiq.log
    :queues:
     - default
     - myqueue

when starting sidekiq service  from the rakefile using "bundle exec sidekiq -r ./sidekiq.rb -C ./sidekiq.yml -d" i see on the log that it loaded the schedules of shcedule.yml and he enqueue them according to the cron arguments.

what i dont fully understand is how flexible the scheduling can be. i want to have the ability to update a job, for example argument or different cron time, and want to add additional jobs while system is running, without the need to restart sidekiq process.

i couldnt find anything working on the net, but i'v found few methods on the sources code which seems should do the job, but still not quite sure how its really should work.

i've tried to use load_scheudle, update_scheudle and reload_schedule! 
for example i've tried to do something like the following for loading additional job dynamicaly

     config =  {
         :class => 'MyWorker',
         :cron => '*/1 * * * * ',
         :args => 'myargs'
     }
     Sidekiq.set_schedule('testSystem1',config)
     Sidekiq.reload_schedule!  # and also tried with Sidekiq::Scheduler.update_schedule
     schedules =Sidekiq.get_all_schedules
     #does show the new schedule but it doesnt really loaded it for actual work of the workers.
i though that i should see on the sidekiq.log something like "Schedule loaded" as it wrote at the startup, and now with the new addition, but it didnt.

using the update_schedule sometimes just give an exception that the class of "@@scheduled_jobs" isnt initalized.

what is the right way to work with Sidekiq-scheduler? 
how i can load when string the sidekiq process few jobs as we do with sidekis.schedule= and continue add additional jobs when the system is running?

any help will be appriciated.

Thanks!

Levi

[...]

PR #67, opened 2015-11-25
Load the schedule dynamically from redis

elpic:
Add the ability to load schedules dynamically from redis on runtime,
the schedules can also be removed dynamically.

* Add the ability to dynamically change schedules.
* Remove an unused capistrano file.


elpic:
Fix #34 and #47

PR #72, opened 2015-12-10
Change enabled configuration text in README

bbauer:
The way I read this, 'disables' should be 'enables'. Is this correct?


elpic:
Already merged

PR #76, opened 2015-12-19
Add "Enqueue now" button

### LINKED ISSUE #75: Add button for manual enqueue jobs through web ui
I currently migrate from sidetiq gem, which now looks unsupported and sometimes behaves unexpected and I concerned by missing buttons for manually starting jobs through web ui. This feature will be very helpful for development.

reneklacan:
screen shot 2015-12-19 at 14 31 55

Closes #75

### DIFF
--- a/lib/sidekiq-scheduler/web.rb
+++ b/lib/sidekiq-scheduler/web.rb
@@ -10,6 +10,12 @@ module SidekiqScheduler
 
         erb File.read(File.join(VIEW_PATH, 'recurring_jobs.erb'))
       end
+
+      app.get '/recurring-jobs/:name/enqueue' do
+        schedule = Sidekiq.get_schedule(params[:name])
+        Sidekiq::Scheduler.enqueue_job(schedule)
+        redirect to('/recurring-jobs')
+      end
     end
   end
 end
--- a/web/views/recurring_jobs.erb
+++ b/web/views/recurring_jobs.erb
@@ -10,6 +10,7 @@
         <th>Class</th>
         <th>Queue</th>
         <th>Arguments</th>
+        <th></th>
       </tr>
     </thead>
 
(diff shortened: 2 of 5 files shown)


web/views/recurring_jobs.erb
```diff
@@ -24,8 +25,13 @@
             <a href="<%= root_path %>queues/<%= job_spec.fetch('queue', 'default') %>"><%= job_spec.fetch('queue', 'default') %></a>
           </td>
           <td><%= job_spec['args'] %></td>
+          <td class="text-center">
+            <a class="btn btn-warn btn-xs" href="<%= root_path %>recurring-jobs/<%= name %>/enqueue">
```
elpic: The name shouldn't be escaped?

reneklacan: urls in href are automatically escaped so even though url includes space it's escaped to `%20` etc

reneklacan:
What could be improved on this one is displaying flash like alert after you hit `Enqueue now` button with message like `SuperJob has been enqueued`

web/views/recurring_jobs.erb
reneklacan: but I will add it just to be 100% sure

elpic:
@reneklacan Sorry for the delay. Could you please squash the commits into one, and rebase with origin? Thanks

lib/sidekiq-scheduler/web.rb
```diff
@@ -10,6 +10,12 @@ def self.registered(app)
 
         erb File.read(File.join(VIEW_PATH, 'recurring_jobs.erb'))
       end
+
+      app.get '/recurring-jobs/:name/enqueue' do
+        schedule = Sidekiq.get_schedule(params[:name])
+        Sidekiq::Scheduler.enqueue_from_config(schedule)
```
elpic: Maybe we can rename this method to enqueue_job. What do you think?

reneklacan: Makes sense :+1:

reneklacan:
@elpic no problem! It's rebased and squashed and I also renamed `enqueue_from_config` to `enqueue_job`

PR #77, opened 2015-12-19
Get rid of huge sleep in tests

reneklacan:
It's really pain in the ass to wait 7 seconds for tests to pass just because of single test. Timecop was already included in dependencies so I utilized it to refactor slow test.


elpic:
Merged

PR #82, opened 2016-01-14
Fix ActiveJob check

sixfeetover:
I was seeing a lot of this in my sidekiq logs:

```
NameError: uninitialized constant ActiveJob::Enqueuing
```

In a Rails, `ActiveJob` is defined even if `ActiveJob::Enqueuing` is not. `ActiveJob::Enqueuing` is only defined if you actually have a class that subclasses `ActiveJob::Base`. This just makes the `defined?` check more explicit.

### DIFF
--- a/lib/sidekiq/scheduler.rb
+++ b/lib/sidekiq/scheduler.rb
@@ -140,7 +140,7 @@ module Sidekiq
       config['class'] = config['class'].constantize if config['class'].is_a?(String)
       config['args'] = Array(config['args'])
 
-      if defined?(ActiveJob) && config['class'].included_modules.include?(ActiveJob::Enqueuing)
+      if defined?(ActiveJob::Enqueuing) && config['class'].included_modules.include?(ActiveJob::Enqueuing)
         config['class'].new.enqueue(config)
       else
         Sidekiq::Client.push(config)


lib/sidekiq/scheduler.rb
```diff
@@ -140,7 +140,7 @@ def self.enqueue_job(job_config)
       config['class'] = config['class'].constantize if config['class'].is_a?(String)
       config['args'] = Array(config['args'])
 
-      if defined?(ActiveJob) && config['class'].included_modules.include?(ActiveJob::Enqueuing)
+      if defined?(ActiveJob::Enqueuing) && config['class'].included_modules.include?(ActiveJob::Enqueuing)
```
elpic: This could be used on non-rails apps. I think this will not work on non-rails app maybe we should add an extra check. What do you think?

sixfeetover: What part wouldn't work in a non-rails app?

elpic: No it is ok. I just tested it on a irb.

elpic:
@sixfeetover could you please rebase this commit. That way I can merge it and include it on a release.

sixfeetover:
Sure, all set.

PR #86, opened 2016-02-06
Localization added

prokopsimek:
Views is possible to localise to foreign languages.
Removed `tilt` gem because of colision with dependencies of gem `sass` in Rails 4 app. Gem looks like unused in `sidekiq-scheduler`.

### DIFF
--- a/web/locales/cs.yml
+++ b/web/locales/cs.yml
@@ -1,6 +1,9 @@
 cs:
-  RecurringJobs: Pravidelně opakované
-  Name: Jméno
-  Description: Popis
-  Interval: Interval
-  EnqueueNow: Zařadit nyní
+  recurring_jobs: Pravidelně opakované
+  name: Jméno
+  description: Popis
+  interval: Interval
+  class: Třída
+  queue: Fronta
+  arguments: Argumenty
+  enqueue_now: Zařadit nyní
--- a/web/locales/cs.yml
+++ b/web/locales/cs.yml
@@ -1,0 +1,6 @@
+cs:
+  RecurringJobs: Pravidelně opakované
+  Name: Jméno
+  Description: Popis
+  Interval: Interval
+  EnqueueNow: Zařadit nyní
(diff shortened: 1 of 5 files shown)


web/locales/cs.yml
```diff
@@ -0,0 +1,6 @@
+cs:
+  'Recurring Jobs': Pravidelně opakované
```
elpic: Could you please use key like strings for this. 
Example:
  recurring_jobs instead of 'Recurring Jobs'.

And add the en.yml. Thanks

prokopsimek: And do you mean `Description` to key like `description` too or not?

prokopsimek:
@elpic changed string key to key like uses Sidekiq original (ref.: https://github.com/mperham/sidekiq/blob/master/web/locales/en.yml)

web/locales/cs.yml
```diff
@@ -0,0 +1,6 @@
+cs:
+  RecurringJobs: Pravidelně opakované
```
elpic: Hi @prokopsimek,

We prefer 'record_invalid'  for 'Record Invalid' if you don't mind.

Thanks

prokopsimek: but I Sidekiq original users camel case, I'd preffer to hold their standards to have some strings (e.g. 'Name') translated from Sidekiq original yaml

web/locales/cs.yml
```diff
@@ -0,0 +1,6 @@
...
+  Name: Jméno
```
elpic: But you also added Name. We follow this guide http://guides.rubyonrails.org/i18n.html.

elpic: Also if they remove the Name translation from his files, we will not have that translation anymore.

prokopsimek: Sorry, I thought `Class` and `Queue`... It's in original yaml already translated... I don't mind to rewrite it, but I think that it would be fine to hold `Class` and `Queue` translated as is in sidekiq repository.

prokopsimek: ok :) i'll do it

prokopsimek:
@elpic done

elpic:
Ok thank you

Issue #91, opened 2016-03-01
Job is scheduled on each queue.


maxrosecollins:
When I schedule a job it is processed on each queue I have.

For example the job will be scheduled and run three times with the following code. Even though it should only run on the `worker` queue it runs on a `worker` queue on both the `reports` and `payment` sidekiqs

Scheduler:
```
:schedule:
  DailyCheckWorker:
    cron: "0 20 00 * * *"
    class: DailyCheckWorker
    queue: worker
    description: "abcdef"
```

Procfile:
```
web: bundle exec puma -C config/puma.rb
reports: bundle exec sidekiq -q reports
worker: bundle exec sidekiq -q worker
payment: bundle exec sidekiq -q payment
```

maxrosecollins:
Is it because `reports` `worker` `payment` are instances of sidekiq running and they each use `sidekiq.rb` which in turn loads the schedule into each instance? 
```
reports: bundle exec sidekiq -q reports
worker: bundle exec sidekiq -q worker
payment: bundle exec sidekiq -q payment
```

elpic:
Hi @maxrosecollins,

Are you sure the jobs is being schedule on each queue? Maybe the message "INFO: queueing Testing (testing)" is the thing that is wrong. 

Because I run a similar example and the jobs only run on his own queue.

maxrosecollins:
Hi,

It definitely queues it on each queue.
For some reason if I do the following it only queues it on the worker queue. The file `./config/sidekiq.rb` doesn't exist by the way... 

```
reports: bundle exec sidekiq -q reports -C ./config/sidekiq.rb
worker: bundle exec sidekiq -q worker
payment: bundle exec sidekiq -q payment -C ./config/sidekiq.rb
```

elpic:
Which version of sidekiq-scheduler are you using?

maxrosecollins:
I'm using `2.0.4`

elpic:
Could you try v2.0.5?

maxrosecollins:
I tried that.
When I start up the server I'm getting

```
11:43:57 mailer.1  | 2016-03-10T11:43:57.660Z 70723 TID-owkmjau4o INFO: Running in ruby 2.2.3p173 (2015-08-18 revision 51636) [x86_64-darwin14]
[...]

PR #93, opened 2016-03-08
do not fail if the scheduled class is not a local class

skippy:
right now, if a class is not in the local path, it cannot be scheduled.

This is not ideal where the scheduler and job are not in the same
project (or where project 1 can enqueue a job that is processed by
project 2)

### DIFF
--- a/lib/sidekiq/scheduler.rb
+++ b/lib/sidekiq/scheduler.rb
@@ -222,7 +222,7 @@ module Sidekiq
     #
     # @return [Boolean]
     def self.active_job_enqueue?(klass)
-      defined?(ActiveJob::Enqueuing) && klass.included_modules.include?(ActiveJob::Enqueuing)
+      defined?(ActiveJob::Enqueuing) && klass.is_a?(Class) && klass.included_modules.include?(ActiveJob::Enqueuing)
     end
 
     # Convert the given arguments in the format expected to be enqueued.
@@ -234,7 +234,9 @@ module Sidekiq
     #
     # @return [Hash]
     def self.prepare_arguments(config)
-      config['class'] = config['class'].constantize if config['class'].is_a?(String)
+      if config['class'].is_a?(String) && Object.const_defined?(config['class'])
+        config['class'] = Object.const_get(config['class'])
+      end
 
       if config['args'].is_a?(Hash)
         config['args'].symbolize_keys! if config['args'].respond_to?(:symbolize_keys!)
--- a/spec/sidekiq/scheduler_spec.rb
+++ b/spec/sidekiq/scheduler_spec.rb
@@ -40,8 +40,29 @@ describe Sidekiq::Scheduler do
         'args'  => '/tmp'
       }
 
-      expect(Sidekiq::Client).to receive(:push).with(process_parameters(config))
+      expect(Sidekiq::Client).to receive(:push) do |opts|
+        expect(opts['class']).to eq(SomeWorker)
+        expect(opts['args']).to eq(['/tmp'])
+      end
 
+      Sidekiq::Scheduler.enqueue_job(config)
+    end
+
+    it 'enqueue does not fail if it cannot constantize' do
+      # The job should be loaded, since a missing rails_env means ALL envs.
+      ENV['RAILS_ENV'] = 'production'
+
+      config = {
+        'cron'  => '* * * * *',
+        'class' => 'SomeOtherWorker',
+        'queue' => 'high',
+        'args'  => '/tmp'
+      }
+
+      expect(Sidekiq::Client).to receive(:push) do |opts|
+        expect(opts['class']).to eq('SomeOtherWorker')
+        expect(opts['args']).to eq(['/tmp'])
+      end
       Sidekiq::Scheduler.enqueue_job(config)
     end
 
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -25,17 +25,4 @@ RSpec.configure do |config|
   config.profile_examples = 10
 
   config.order = :random
-
-  def process_parameters(config)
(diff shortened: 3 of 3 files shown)


lib/sidekiq/scheduler.rb
```diff
@@ -137,7 +137,9 @@ def self.handle_errors
     def self.enqueue_job(job_config)
       config = job_config.dup
 
-      config['class'] = config['class'].constantize if config['class'].is_a?(String)
+      if config['class'].is_a?(String) && Object.const_defined?(config['class'])
+        config['class'] = config['class'].constantize
```
snmgian: `#constantize` is part of ActiveSupport API, so `config['class'].constantize` will fail on applications/projects which are not relying on ActiveSupport. It seems that `Kernel#const_get` would do the job.

lib/sidekiq/scheduler.rb
```diff
@@ -137,7 +137,9 @@ def self.handle_errors
...
+      end
       config['args'] = Array(config['args'])
 
       if defined?(ActiveJob::Enqueuing) && config['class'].included_modules.include?(ActiveJob::Enqueuing)
```
snmgian: If 'project 1' depends on ActiveJob but the job's class exists only in 'project 2', `config['class'].included_modules` will fail. I think that config['class'] type should be verified before asking for the included modules.

snmgian:
I just posted some comments, please feel free to discuss them if you think is appropriate. Thanks for the pull request!

spec/sidekiq/scheduler_spec.rb
```diff
@@ -40,8 +40,29 @@
         'args'  => '/tmp'
       }
 
-      expect(Sidekiq::Client).to receive(:push).with(process_parameters(config))
+      expect(Sidekiq::Client).to receive(:push) do |opts|
+        expect(opts['class']).to eq(SomeWorker)
+        expect(opts['args']).to eq(['/tmp'])
```
skippy: just making it explicit... I actually couldn't tell what was being tested.  Hope you don't mind :)

spec/sidekiq/scheduler_spec.rb
```diff
@@ -40,8 +40,29 @@
...
+        'cron'  => '* * * * *',
+        'class' => 'SomeOtherWorker',
+        'queue' => 'high',
+        'args'  => '/tmp'
+      }
+
+      expect(Sidekiq::Client).to receive(:push) do |opts|
+        expect(opts['class']).to eq('SomeOtherWorker')
+        expect(opts['args']).to eq(['/tmp'])
+      end
```
skippy: so this is the use case... since there is one universal schedule, sometimes a scheduler that is in the context of project1 will schedule a job from project2; this will test that it won't fail.  

This project doesn't need to constantize the class for any reason except for ActiveJob, which I have another comment on.

lib/sidekiq/scheduler.rb
```diff
@@ -234,7 +234,9 @@ def self.active_job_enqueue?(klass)
     #
     # @return [Hash]
     def self.prepare_arguments(config)
-      config['class'] = config['class'].constantize if config['class'].is_a?(String)
+      if config['class'].is_a?(String) && Object.const_defined?(config['class'])
+        config['class'] = Object.const_get(config['class'])
+      end
```
skippy: I'm not really a fan of this as the behavior silently changes on whether the object exists in the load path (never mind if it is auto_loaded!)  My challenge is I don't know ActiveJob... I did some tracing and it definitely doesn't look like it is as 'simple' as not instantiating the class and just passing a hash to get queued by Sidekiq!  

Some more explicit options may be to:
* make an explicit configuration flag, such as `Sidekiq::Scheduler.enable_active_job = true` (default to true)
* print a `logger.warn` msg if the class cannot be constantized.

Any suggestions on how to thread this needle?

snmgian: It seems a good idea to have an explicit flag. Its default value would be `true` if  `defined?(ActiveJob::Enqueing)`. 

Also, it would be great to be able to configure that flag at job level, because activejobs and non-activejobs could coexist in the same project.

Finally, in order to be able to schedule an ActiveJob from project1 and be executed in project2, that job has to be pushed into sidekiq by following the same idea as `ActiveJob::QueueAdapters::SidekiqAdapter`:

```
        {
          'class'        => 'ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper',
          'queue'        => config['queue'],
          'description'  => config['description'],
          'args'         => [{
            'job_class'  => config['class'],
            'job_id'     => SecureRandom.uuid,
            'queue_name' => config['queue'],
            'arguments'  => config['args']
          }]
        }
```

Any thoughts?

snmgian:
Hi @skippy,

Recently, a change was made in order to not schedule jobs whose queues are not being listened on by the current sidekiq process (PR: https://github.com/moove-it/sidekiq-scheduler/pull/96). That change was made in order to avoid some confusion (Issue: https://github.com/moove-it/sidekiq-scheduler/issues/91)

That change will conflict with the feature you are working on, because if the worker class it not a local one, the local sidekiq process can't be listening on the scheduled job's queue; if the queue is not enabled, the scheduler will not enqueue the job.

A way to solve this would be to have some config flag that marks the job as _remote_, and then don't ignore schedules for those remote jobs. 

```yml
cron: '* * * * *'
class: 'SomeOtherWorker'
remote: true
queue: 'high'
args: '/tmp'
```

I think that this flag could help to resolve some problems we were discussing about class loading. It will also don't fail silently in the case of the job's class doesn't exist due to some programming error.

What do you think? 

If for some reason you are not able to allocate the required programming effort, I could give you a hand. I think the feature of scheduling non local classes is a great one to have :)

skippy:
thanks folks; I appreciate it!

Issue #102, opened 2016-04-08
When schedules names are of type Symbol, they are removed from redis in the next process start


snmgian:
When `sidekiq.yml` has the following:

```yml
:schedule:
  :remove_temp_files: # This key is loaded as a Symbol
    cron: " ..." 
```

The schedule `:remove_temp_files` is stored in Redis during sidekiq startup, while during the next startup is removed from redis.  In the next startup the shedule will stored in Redis.

This occurs because `SidekiqScheduler::Schedule#get_all_schedules` fetches a `Hash` of schedules from Redis, and that hash's keys are strings.

The fetch/remove/add occurs here:
 https://github.com/moove-it/sidekiq-scheduler/blob/master/lib/sidekiq-scheduler/schedule.rb#L40

PR #108, opened 2016-05-04
Fix bug when job key is a symbol and sidekiq runs twice

matu1104:
This change is related to https://github.com/moove-it/sidekiq-scheduler/issues/102

### DIFF
--- a/spec/sidekiq-scheduler/schedule_spec.rb
+++ b/spec/sidekiq-scheduler/schedule_spec.rb
@@ -1,4 +1,5 @@
 describe SidekiqScheduler::Schedule do
+  before { Sidekiq.redis(&:flushall) }
 
   def build_cron_hash
     {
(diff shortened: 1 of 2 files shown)


spec/sidekiq-scheduler/schedule_spec.rb
```diff
@@ -29,24 +31,34 @@ def job_from_redis_without_decoding(job_id)
...
 
       expect(cron_hash).to eq(job_from_redis(job_class_id))
       expect(job['class']).to eq(job_class_id)
     end
+
+    context 'when job key is a symbol' do
+      let(:job_id) { :super_job }
+
+      context 'when schedule is set twice' do
+        it 'sets the schedule on redis' do
```
snmgian: This test does not fail when running the specs with seed = 9838. Please flush Redis on a `before` hook (see `spec/sidekiq/scheduler_spec.rb`)

matu1104: Thanks for the comment. I just pushed a new change.

PR #114, opened 2016-07-05
Specify ActiveJob < 5 when RUBY_VERSION < 2.2.2

snmgian:
ActiveJob 5 requires ruby >= 2.2.2, so we need to specify an older version when
ruby version requirement is not met.


cdotta:
👍

PR #118, opened 2016-07-25
Update readme

mihai-dinculescu:
It should go 2 folders up, not 3. 
It doesn't error out, but it's confusing nonetheless.


mihai-dinculescu:
Closing and redoing because I've failed at rebase.

PR #130, opened 2016-09-14
Enqueue workers whose class is not available

snmgian:


lib/sidekiq/scheduler.rb
```diff
@@ -258,7 +258,9 @@ def initialize_active_job(klass, args)
       #
       # @return [Boolean]
       def active_job_enqueue?(klass)
-        defined?(ActiveJob::Enqueuing) && klass.included_modules.include?(ActiveJob::Enqueuing)
+        klass.is_a?(Class) &&
+          defined?(ActiveJob::Enqueuing) &&
+          klass.included_modules.include?(ActiveJob::Enqueuing)
```
cdotta: indent it as the previous one

PR #134, opened 2016-09-21
Stringify configs

phongsi:
currently sidekiq expects symbols in their yml file. chef cookbooks also expect this format and converts it to match. example chef cookbook that currently matches this behaviour https://supermarket.chef.io/cookbooks/opsworks_ruby

However, sidekiq-scheduler only expects the highest level :schedule: and the rest of the config is read through strings. this fix ensures that using sidekiq format also works, and the current chef cookbooks will also work.

### DIFF
--- a/lib/sidekiq-scheduler.rb
+++ b/lib/sidekiq-scheduler.rb
@@ -21,7 +21,7 @@ Sidekiq.configure_server do |config|
     scheduler_options = {
       dynamic:   dynamic,
       enabled:   enabled,
-      schedule:  config.options.fetch(:schedule, nil),
+      schedule:  config.options.fetch(:schedule, nil).deep_stringify_keys,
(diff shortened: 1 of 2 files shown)


snmgian:
Hi @phongsi 

Adding `activesupport ~> 5` as a dependency will prevent other users who are locking in `activesupport < 5` to use sidkeiq-scheduler. 

ActiveSupport is a huge library that adds a lot of methods to the core Ruby clases, so people who are not relying in ActiveSupport could suddenly face some compatibility problems.

Said that, I'll create an issue so we can afford the need in some other way :)

snmgian:
Of course, I'm open to discussion ;)

snmgian:
Isse #135 created.

phongsi:
wow thanks for that, i can close this pr, and discuss it in the issue thread.

Issue #135, opened 2016-09-21
Symbolize config keys


snmgian:
See PR #134 

Find a way that don't depend on ActiveSupport

> currently sidekiq expects symbols in their yml file. chef cookbooks also expect this format and converts it to match. example chef cookbook that currently matches this behaviour https://supermarket.chef.io/cookbooks/opsworks_ruby

> However, sidekiq-scheduler only expects the highest level :schedule: and the rest of the config is read through strings. this fix ensures that using sidekiq format also works, and the current chef cookbooks will also work.

phongsi:
I've been looking over some options that provides the functionality

Hashie might provide a lot more for what I need, but it might make the code cleaner if you use what it provides.

https://github.com/intridea/hashie
https://rubygems.org/gems/hashwithindifferentaccess/versions/0.1.1

phongsi:
i ended up using just the hash ext from activesupport wrapped into a gem of itself.

https://github.com/LightshedHealth/sidekiq-scheduler/commit/3a91612aa3ac894d41b3be6243782313265b5044

snmgian:
If sidekiq.yml has the following content:
```yaml
:schedule:
  :interval:
    :interval: "20s"
    :class: HelloWorker
    :description: "abcdef"
```
When sidekiq starts up it outputs this and don't schedules any job.
```
2016-10-01T01:49:39.317Z 26437 TID-ov1dze5aw INFO: Scheduling interval {:interval=>"20s", :class=>"HelloWorker", :description=>"abcdef"}
2016-10-01T01:49:39.317Z 26437 TID-ov1dze5aw INFO: no cron / every / at / in / interval found for  (interval) - skipping
```

Should'n the job be scheduled as if the config is made of strings only?

snmgian:
See issue #140

PR #138, opened 2016-09-26
add hash with indifferent access

phongsi:
referencing [issue #135](https://github.com/moove-it/sidekiq-scheduler/issues/135)
This gem is just the hash ext from active support and nothing else.

It adds the least amount of 3rd party code, and does it nicely.

The one test was modified cause it was basically pushing a new key to the options hash rather than replacing the existing key. This is one of those problems with adding a symbol key to a hash with string keys. But with hashwithindifferent access, that problem goes away cause the keys are treated as if they were the same.

This also allows the yml configs to be symbols or strings.


snmgian:
Thanks!

PR #151, opened 2016-10-25
Set schedule if it was not previously set

snmgian:
On startup sidekiq hook, set the schedule if it was not previously set,
following the same mechanism as other configuration flags.

### DIFF
--- a/spec/sidekiq-scheduler/schedule_spec.rb
+++ b/spec/sidekiq-scheduler/schedule_spec.rb
@@ -96,10 +96,8 @@ describe SidekiqScheduler::Schedule do
   describe '.get_schedule' do
     subject { Sidekiq.get_schedule(job_id) }
 
-    context 'when schedules previously set' do
-      before do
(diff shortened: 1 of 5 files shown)


spec/sidekiq-scheduler/schedule_spec.rb
```diff
@@ -94,12 +94,30 @@ def job_from_redis_without_decoding(job_id)
   end
 
   describe '.get_schedule' do
-    it 'returns a schedule' do
-      Sidekiq.set_schedule(job_id, cron_hash)
+    subject { Sidekiq.get_schedule(job_id) }
+
+    context 'when schedules previously set' do
```
cdotta: when schedules are previously set

spec/sidekiq-scheduler/schedule_spec.rb
```diff
@@ -94,12 +94,30 @@ def job_from_redis_without_decoding(job_id)
...
+      before do
```
cdotta: one line before block is too long?

PR #158, opened 2016-12-21
change min rufus-scheduler dep to ~> 3.2

cih:
Firstly thanks for your work on this project!

We're starting the process of moving from resque to sidekiq and will be running the two side-by-side while we complete the transition.

resque-scheduler has a rufus-scheduler dependency on `~> 3.2`  while sidekiq-scheduler currently has `~> 3.1.8`. Would you be happy to bump or relax the version constraints?

https://github.com/resque/resque-scheduler/blob/master/resque-scheduler.gemspec#L59

https://github.com/jmettraux/rufus-scheduler/compare/v3.1.8...v3.2.0?expand=1


snmgian:
Thanks

PR #159, opened 2016-12-27
Fixed typo

rnbd:
In the 'Loading the schedule' section it was assumed the schedule filed was named 'scheduler.yml', but the initializer was set to load 'sidekiq_scheduler.yml'.


snmgian:
Thanks

Issue #157, opened 2016-12-19
Loading queue from Sidekiq options


cih:
Would you be open to a PR that checks the worker options to determine the queue name in addition to checking config file and falling back to "default"? If so I would be happy to make this change.

snmgian:
Sure!

Do you plan to handle ActiveJob workers as well? See http://edgeguides.rubyonrails.org/active_job_basics.html#queues

cih:
Cool I'll take a look into this then, yep I'll make sure to handle ActiveJob workers too.

PR #160, opened 2016-12-28
Infer queue name when loading schedule

cih:
Fixes https://github.com/moove-it/sidekiq-scheduler/issues/157

This change updates the `SidekiqScheduler::Schedule#schedule=` to
infer queue name from the worker. It supports both sidekiq workers
and active jobs by checking for `sidekiq_options` and a `queue_name`.

Note: This change does not update the `set_schedule` method to infer
the queue name but could be updated to handle this case if required.


snmgian:
Thanks.

A new version will be released at end of week.

PR #161, opened 2016-12-29
Fix error JSON::ParserError: lexical error: invalid char in json text on schedule loading

zzet:
We send `hash` as param into `set_schedule`, not json (string). In this case `JSON(hash)` raised exception (`JSON::ParserError: lexical error: invalid char in json text.`) and we do not need parse this arg.

### DIFF
--- a/lib/sidekiq-scheduler/schedule.rb
+++ b/lib/sidekiq-scheduler/schedule.rb
@@ -102,7 +102,7 @@ module SidekiqScheduler
     def set_schedule(name, config)
       existing_config = get_schedule(name)
       unless existing_config && existing_config == config
-        Sidekiq.redis { |r| r.hset(:schedules, name, JSON(config)) }
+        Sidekiq.redis { |r| r.hset(:schedules, name, JSON.generate(config)) }
         Sidekiq.redis { |r| r.sadd(:schedules_changed, name) }
       end
       config


snmgian:
Hi @zzet,

`set_schedule` already works with Hashes, so doing smth like the following should already work:

``` ruby
Sidekiq.set_schedule('HelloWorld', { cron: '*/15 * * * * *', class: 'HelloWorld' })
```

Do you have a chance to provide the backtrace you were facing? Thanks

zzet:
```
Sidekiq.set_schedule('HelloWorld', { cron: '*/15 * * * * *', class: 'HelloWorld' })
JSON::ParserError: parse error: invalid object key (must be a string)
                                      {:cron=>"*/15 * * * * *", :class
                     (right here) ------^
from /bundle/gems/yajl-ruby-1.3.0/lib/yajl/json_gem/parsing.rb:15:in `rescue in parse'
```

zzet:
```ruby
# Sidekiq.set_schedule('HelloWorld', { cron: '*/15 * * * * *', class: 'HelloWorld' })
def set_schedule(name, config)
  # name == 'HelloWorld'
  # name.class == String
  # config == { cron: '*/15 * * * * *', class: 'HelloWorld' }
  # config.class == Hash

  existing_config = get_schedule(name)
  # existing_config == nil

  unless existing_config && existing_config == config 
    # condition == true
    Sidekiq.redis { |r| r.hset(:schedules, name, JSON(config)) }
    # JSON(config) == JSON.parse(config) == JSON.parse(Hash.new(config)) == Exception (Hash != String)
    # Sidekiq.redis { |r| r.hset(:schedules, name, JSON.dump(config)) } fix this
    Sidekiq.redis { |r| r.sadd(:schedules_changed, name) }
  end
  config
end
```

snmgian:
I noticed that you are using 'yajl' gem. That gem seems not to provide the same API as the standard JSON implementarion of Ruby.

``` ruby
require 'json'
JSON({ cron: '*/15 * * * * *', class: 'HelloWorld' })

# returns this string: {"cron":"*/15 * * * * *","class":"HelloWorld"}
```

``` ruby
require 'yajl/json_gem'
JSON({ cron: '*/15 * * * * *', class: 'HelloWorld' })

JSON::ParserError: parse error: invalid object key (must be a string)
```

0) Do you know if `yagl` gem plans to support the same interface as `Kernel#JSON` ?
1) Could you replace the other references to `#JSON` -> `JSON.parse`
2) Just for sake of curiosity, why do you choose `JSON.dump` instead of `JSON.generate` ?

zzet:
@snmgian 👍

PR #165, opened 2017-01-13
Rewrite README

snmgian:


README.md
```diff
@@ -25,269 +25,221 @@
...
-  config.on(:startup) do
-    Sidekiq.schedule = YAML.load_file(File.expand_path('../../scheduler.yml', __FILE__))
-    Sidekiq::Scheduler.reload_schedule!
-  end
-end
+See https://github.com/jmettraux/rufus-scheduler for more information.
+
+## Load the schedule from a different file
+
+You can place the schedule configuration in a separate file than `config/sidekiq.yml`
```
migl: "... place the schedule configuration in a separate file **from** `config/sidekiq.yml`"

README.md
```diff
@@ -25,269 +25,221 @@
...
+clear_leaderboards_contributors:
+  cron: '0 30 6 * * 1'
+  class: ClearLeaderboards
+  queue: low
+  args: contributors
+  description: 'This job resets the weekly leaderboard for contributions'
 ```
 
-If you are running a non Rails project you should add code to load the workers classes before loading the schedule.
+Please note that `schedule` root key is not present in the separate file.
```
migl: "Please **notice** that **the** `schedule` root key is not present..."

README.md
```diff
@@ -25,269 +25,221 @@
...
 
-``` ruby
-# config/initializers/sidekiq_scheduler.rb
-require 'sidekiq/scheduler'
+## Configuration options
 
-puts "Sidekiq.server? is #{Sidekiq.server?.inspect}"
-puts "defined?(Rails::Server) is #{defined?(Rails::Server).inspect}"
-puts "defined?(Unicorn) is #{defined?(Unicorn).inspect}"
+Configuration options are placed inside `sidekiq.yml` config file.
```
migl: I'd clarify that they are placed here by default so that it leads to the part where you explain how to use a different config file.

snmgian: These options can only be placed in sidekiq config file. They can not be placed in a separate file.

PR #168, opened 2017-01-26
Fix 'next time' when schedule includes options

snmgian:
Display right 'next time' values in Web UI

When options in form of array were passed to the schedule config, 'next time' column
was not being updated.

Schedule configuration sample causing the bug:

```yaml
:schedule:
  hello_world:
    every: ['30s', first_in: '120s']
    class: HelloWorld
```

### DIFF
--- a/lib/sidekiq-scheduler/rufus_utils.rb
+++ b/lib/sidekiq-scheduler/rufus_utils.rb
@@ -1,0 +1,29 @@
+require 'hashie'
+
+module SidekiqScheduler
+  class RufusUtils
+
+    # Normalizes schedule options to rufust scheduler options
+    #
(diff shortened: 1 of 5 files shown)


lib/sidekiq-scheduler/rufus_utils.rb
```diff
@@ -0,0 +1,29 @@
...
+  class RufusUtils
+
+    # Normalizes schedule options to rufust scheduler options
+    #
+    # @param options [String, [Array]
+    #
+    # @return [Array]
+    #
+    # @example
+    #   normalize_schedule_options('15m') => ['12d', {}]
```
cdotta: 15m 15m

PR #169, opened 2017-02-01
Use precise log message when updating schedule

Tolsto:
update_schedule should use a different log message than load_schedule!


snmgian:
Thanks!

PR #171, opened 2017-02-03
Don't set schedule when not enabled

cih:
This change updates the manager to only set the schedule when
the `enabled` options is `true`. This means that we wont update 
the schedule in Redis when the config passes `enabled` as `false`.

The use case for this was running two processes with different
config options eg.

`bundle exec sidekiq -C config/sidekiq.yml`
`bundle exec sidekiq -C config/sidekiq_schedule.yml`

The `sidekiq_schedule.yml` contains a collection of scheduled
jobs and the `sidekiq.yml` does not contain any scheduled jobs.
What we were seeing was that depending on which is last to run
the `Sidekiq.get_schedule` (from Redis) could either contain the
schedule or return an empty hash.

### DIFF
--- a/lib/sidekiq-scheduler/manager.rb
+++ b/lib/sidekiq-scheduler/manager.rb
@@ -18,7 +18,7 @@ module SidekiqScheduler
       Sidekiq::Scheduler.enabled = options[:enabled]
       Sidekiq::Scheduler.dynamic = options[:dynamic]
       Sidekiq::Scheduler.listened_queues_only = options[:listened_queues_only]
-      Sidekiq.schedule = options[:schedule]
+      Sidekiq.schedule = options[:schedule] if Sidekiq::Scheduler.enabled
     end
 
     def stop
--- a/spec/sidekiq-scheduler/manager_spec.rb
+++ b/spec/sidekiq-scheduler/manager_spec.rb
@@ -1,5 +1,4 @@
 describe SidekiqScheduler::Manager do
-
   describe '.new' do
     let(:previous_schedule) do
       {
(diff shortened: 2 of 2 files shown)


humzashah:
👍 really need this

question: would this still respect the default `enabled` value of `true`? i.e. we wouldn't have to explicitly set it to `true` for it to be enabled?

cih:
@humzashah yes that's right it would continue to respect the defaults here https://github.com/moove-it/sidekiq-scheduler/blob/master/lib/sidekiq-scheduler.rb#L13-L14

lib/sidekiq-scheduler/manager.rb
```diff
@@ -18,21 +18,28 @@ def initialize(options)
...
 
     def stop
       Sidekiq::Scheduler.clear_schedule!
     end
 
     def start
-      Sidekiq::Scheduler.load_schedule!
+      if enabled
+        Sidekiq.schedule = schedule
+        Sidekiq::Scheduler.load_schedule!
```
snmgian: Thanks for the PR.

Sidekiq::Scheduler#load_schedule! checks for `enabled` flag, so just to avoid repetion, how about removing `.enabled=` from `Sidekiq::Scheduler` and also remove the `if` structure in that class?

cih: I think I can safely remove the conditional from the `Sidekiq::Scheduler#load_schedule!` but I think the enabled needs to remain for the `reload_schedule!` method?

snmgian: You're right, that flag is needed by `reload_schedule!`

How about conditionally invoking `Sidekiq.schedule = ...` if enabled? Doing that way `#load_schedule!` we'll still have an extra conditional but will not need for extra instance variables, and it will also log `'SidekiqScheduler is disabled'` message.

```ruby
    def initialize(options)
      # ...
      Sidekiq.schedule = options[:schedule] if Sidekiq::Scheduler.enabled
    end
```
Just an idea :)

snmgian: By making this change to `#initialize`, the change in `#start` will not be needed

cih: Yep that seems reasonable to me, will update

snmgian:
Thanks!

PR #176, opened 2017-03-27
Config hashie object with indifferent access

hunzinker:
Thanks for the great gem!

I noticed a bug where `ActiveJob` scheduled jobs do not recognize jobs scheduled with a `queue` key value.

ENV:
```
    rails (4.2.7.1)
    sidekiq (4.2.10)
    sidekiq-pro (3.4.5)
    sidekiq-scheduler (2.1.2)
```

Example Scheduler:
```
 job_name:
    cron: "30 0 * * * America/Los_Angeles"
    class: WorkerClass
    args: ["args", "for", "worker"]
    queue: "host-specific-queue"
    description: "process stuff"
```
 
The scheduled job above, is parsed as a hash with stringified keys. When the hash is passed to `WorkerClass.enqueue(config)` the stringified keys are ignored and the job is enqueued on the `default` queue unless `queue_as` is set.

Solution:
The gem is already including `Hashie` so I propose converting the config hash to a hashie hash with indifferent access before enqueuing.

I have not tested this change in Rails 5 so it's unclear if the hash with indifferent access is needed.

Current Workaround:
To avoid this issue, I added the below:

```ruby
class ApplicationJob < ActiveJob::Base
  def enqueue(options = {})
    super(options.symbolize_keys)
  end
end
```

This PR is intended as a example solution to open dialogue.

Thanks!

### DIFF
--- a/lib/sidekiq-scheduler/config_hash.rb
+++ b/lib/sidekiq-scheduler/config_hash.rb
@@ -1,0 +1,8 @@
+require 'hashie'
+
+module SidekiqScheduler
+  class ConfigHash < Hash
+    include Hashie::Extensions::MergeInitializer
+    include Hashie::Extensions::IndifferentAccess
+  end
+end
--- a/lib/sidekiq/scheduler.rb
+++ b/lib/sidekiq/scheduler.rb
@@ -3,6 +3,7 @@ require 'thwait'
 require 'sidekiq/util'
 require 'sidekiq-scheduler/manager'
 require 'sidekiq-scheduler/rufus_utils'
+require 'sidekiq-scheduler/config_hash'
 require 'json'
 
 module Sidekiq
@@ -266,6 +267,8 @@ module Sidekiq
       #
       # @return [Hash]
       def prepare_arguments(config)
+        config = SidekiqScheduler::ConfigHash.new(config)
+
         config['class'] = try_to_constantize(config['class'])
 
         if config['args'].is_a?(Hash)


snmgian:
Hi @hunzinker, thanks for finding and tracking down this.

ActiveJob instances receive these options through method `#enqueue`:
 - http://api.rubyonrails.org/classes/ActiveJob/Enqueuing.html#method-i-enqueue
 - http://api.rubyonrails.org/v4.2.3/classes/ActiveJob/Enqueuing.html#method-i-enqueue

So, instead of having a hash with indifferent access, why not simply build that options hash?

Doing that will also prevent us to pass to `#enqueue` the other options like `class` and `args`, though it seems that there's no problem passing those extra options, I think its better to stick to `#enqueue` documentation.

I pushed another PR with that proposal: PR #177 

I've tested that approach in Rails v4.2.3 and Rails v5.0.2

Thoughts?

hunzinker:
@snmgian thanks for the feedback!

After reviewing your PR I noticed one concern with https://github.com/moove-it/sidekiq-scheduler/pull/177/files#diff-d5f01781450b17c68db7c73404edd78bR233

Why limit to only passing `enqueue`? It's possible a user would want to provide `priority` via the `sidekiq_scheduler.yml` as well? When Rails 6 is released some awesome new option `key => value`.

My proposal, converting the options hash to a hash with indifferent access, covers having to make specific changes for sidekiq and active_job should they arise.

It boils down to how you want to handle the problem in the end. That's where I'll defer to you since you're the expert. :)

snmgian:
Allowing users to pass in options to `#enqueue` is a nice feature. But those options should be specified at a different level in the configuration hierarchy to prevent any collision with former schedule's options.

What if in Rails 6 an option named `include_metadata` is added? What if we want to  add a `wait` config option with different semantics than `#enqueue`'s wait option? What if in a future version of ActiveJob it starts to emit warnings when unkown options are passed in?

Those issues will be prevented by grouping options in a separate key. This is an example of such configuration:

``` yaml
 job_name:
    cron: "30 0 * * * America/Los_Angeles"
    class: WorkerClass
    args: ["args", "for", "worker"]
    queue: "host-specific-queue"
    include_metadata: true
    wait: true # Wait any previous job to finish
    description: "process stuff"
    actvejob: # ActiveJob specific options
      wait: 5.minutes
      include_metadata: { host: 'localhost' }
```

snmgian:
PR #177 merged

snmgian:
v2.1.3 released

PR #179, opened 2017-04-19
German localization

trautwein:
Thanks so much for developing this gem. 👍 

I added a german locale file in this PR. Would be great if you'd merge it.


snmgian:
Thanks!

PR #187, opened 2017-05-15
Fix logo's src in README.md

matu1104:


snmgian:
Changed img's src to 'https://moove-it.github.io/sidekiq-scheduler/images/small-logo.svg'. See #189

Issue #191, opened 2017-06-05
Suggestion: add a option to disable remove existing jobs when initializing


IvanChou:
I have two project which use the same redis db and both have sidekiq-scheduler. Let's call them `ProjectA` and `ProjectB`

ProjectA has a scheduler named `project_a_cron_task`

ProjectB has a scheduler named `project_b_cron_task`

When I start ProjectA (sidekiq only), everything is fine, I can find `project_a_cron_task` on sidekiq web page.

Then, I start ProjectB (sidekiq only),  there is only `project_b_cron_task` on the web page, `project_a_cron_task` is disappeared! WHAT HAPPEND?

After reading the sources, I find `Sidekiq#schedule=` will be called everytime when we start/restart a sidekiq with sidekiq-scheduler.

In `#schedule =`

```ruby
    def schedule=(schedule_hash)
      schedule_hash = prepare_schedule(schedule_hash)
      to_remove = get_all_schedules.keys - schedule_hash.keys.map(&:to_s)

      schedule_hash.each do |name, job_spec|
        set_schedule(name, job_spec)
      end

      to_remove.each do |name|
        remove_schedule(name)
      end

      @schedule = schedule_hash
    end
```

The `schedule_hash` is options of current schedules loading from `redis.yml`. When ProjectB starting, it's options of `project_b_cron_task`. 

`get_all_schedules` return existing jobs from Redis, `project_a_cron_task` is contained.

So, `project_a_cron_task` is removed from `schedules` (it's a hash in Redis) and disapperd from web page. 

If you don't set `dynamic: ture`, `project_a_cron_task` still will be ran correctly. (yeah, I can't see it, but it's still here~)

If you set `dynamic: ture`, there will be more complicated, because when :dynamic flag is set to true, schedule changes are loaded every 5 seconds, `remove_schedule` is one of schedule changes.

schedule changes are loaded by `Sidekiq::Scheduler#update_schedule` which uses `spop` from 'schedules_changed'(set) to check what to reload.

**!! SPOP !!**, that means only one instance can load the changes, others don't know what has happend.

After ProjectB starting, 'schedules_changed' has been updated, if

- ProjectA has only one instance and triggers `update_schedule` firstly,  `project_a_cron_task` will be stoped.
- ProjectA has multiple instances, no matter who triggers `update_schedule` firstly, `project_a_cron_task` will still be ran
- ProjectB triggers `update_schedule` firstly, `project_a_cron_task` will still be ran

[...]

PR #193, opened 2017-06-11
Use Sorted-Set instead of Set to store 'schedules_changed' in redis

IvanChou:
As we discussed in issue #191 , `spop` may lead to unexpected results under distributed scene.

In this PR, I use `Sorted-Set` instead of `Set` to store schedules_changed info. It works like the db migration of Rails.

Notice, This RP will lead a break change:

If a schedules_change has been added before sidekiq server client starts up, the schedules_change will be ignored by the new instance of sidekiq server. Because the member of `schedules_changed` can only be update (its score) and never be deleted, we can't judge which is outdated or pre-added.

### DIFF
--- a/lib/sidekiq-scheduler/schedule.rb
+++ b/lib/sidekiq-scheduler/schedule.rb
@@ -104,7 +104,7 @@ module SidekiqScheduler
       existing_config = get_schedule(name)
       unless existing_config && existing_config == config
         Sidekiq.redis { |r| r.hset(:schedules, name, JSON.generate(config)) }
-        Sidekiq.redis { |r| r.sadd(:schedules_changed, name) }
+        Sidekiq.redis { |r| r.zadd(:schedules_changed, Time.now.to_f, name) }
       end
       config
     end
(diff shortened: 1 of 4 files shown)


snmgian:
Hi @IvanChou, thanks for the PR. 

I've been thinking a while about an issue here and finally after not arriving to a good answer I want to share it with you ...

`schedules_changed` sorted set contents are never flushed, so theoretically it can grow and become a very large sized set. How can we prevent this while at the same time be sure changes will be reflected on multiple instances?

Thought?

Sorry for the slow response ...

IvanChou:
I think it is almost impossible to grow and become a very large sized set.

Because

> Sorted sets are a data type which is similar to a mix between a Set and a Hash. Like sets, sorted sets are composed of unique, non-repeating string elements, so in some sense a sorted set is a set as well.

Maybe we will change our schedules frequently, but the total number of schedules is sure. The max size of the `schedules_changed` can't be greater than the total number of schedules.

The complexity of `ZRANGEBYSCORE` is O(log(N)+M), it's acceptable, I think.

snmgian:
This will be part of the next release, thanks!

PR #197, opened 2017-06-19
Refactor scheduler tests

mconiglio:
Refactored the following tests:

- job_presenter_spec
- manager_spec
- rufus_utils_spec
- schedule_spec
- utils_spec
- web_spec

### DIFF
--- a/spec/sidekiq-scheduler/job_presenter_spec.rb
+++ b/spec/sidekiq-scheduler/job_presenter_spec.rb
@@ -2,6 +2,7 @@ require 'sidekiq-scheduler/job_presenter'
 
 describe SidekiqScheduler::JobPresenter do
   subject(:job_presenter) { described_class.new(job_name, attributes) }
+
   let(:job_name) { 'job_name' }
   let(:attributes) { {} }
 
@@ -10,12 +11,15 @@ describe SidekiqScheduler::JobPresenter do
   describe '#next_time' do
     subject { job_presenter.next_time }
 
+    before { Sidekiq::Scheduler.update_job_next_time(job_name, next_time) }
+
     context "when the job doesn't have a next time in redis" do
+      let(:next_time) { nil }
+
       it { is_expected.to be_nil }
     end
 
     context 'when the job has a next time in redis' do
-      before { Sidekiq::Scheduler.update_job_next_time(job_name, next_time) }
       let(:next_time) { Time.now }
 
       it { is_expected.to eq(job_presenter.relative_time(next_time)) }
@@ -25,22 +29,22 @@ describe SidekiqScheduler::JobPresenter do
   describe '#interval' do
     subject { job_presenter.interval }
 
-    let(:attributes) { { 'cron' => 'cron_value' } }
+    context 'with "cron" key' do
+      let(:attributes) { { 'cron' => 'cron_value' } }
 
-    it { is_expected.to eq('cron_value') }
+      it { is_expected.to eq('cron_value') }
+    end
 
-    context "when the attributes don't have a cron key" do
-      context 'when the attributes have an interval key' do
-        let(:attributes) { { 'interval' => 'interval_value' } }
+    context 'with "interval" key' do
+      let(:attributes) { { 'interval' => 'interval_value' } }
 
-        it { is_expected.to eq('interval_value') }
-      end
+      it { is_expected.to eq('interval_value') }
+    end
 
-      context 'when the attributes have an every key' do
-        let(:attributes) { { 'every' => 'every_value' } }
+    context 'with "every" key' do
+      let(:attributes) { { 'every' => 'every_value' } }
 
-        it { is_expected.to eq('every_value') }
-      end
+      it { is_expected.to eq('every_value') }
     end
   end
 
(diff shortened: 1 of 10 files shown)


spec/sidekiq-scheduler/job_presenter_spec.rb
```diff
@@ -1,116 +1,99 @@
...
-      end
+      it { is_expected.to eq(job_presenter.relative_time(next_time)) }
     end
   end
 
-  describe "#interval" do
-    context "when the attributes has a cron key" do
-      let(:attributes) { { "cron" => cron_value } }
-      let(:cron_value) { "cron_value" }
+  describe '#interval' do
```
snmgian: There's no base case for these set of tests; either interval, nor every are edge cases.

So, in this situation, it would be ok to have smth like:

``` ruby
context 'with "cron" key' do
  // ...
end

context 'with "interval" key' do
  // ...
end

context 'with "every" key' do
  // ...
end
```

spec/sidekiq-scheduler/manager_spec.rb
```diff
@@ -19,48 +16,25 @@
...
+      let(:enabled) { true }
 
-    it 'sets Sidekiq.schedule' do
-      expect {
-        subject
-      }.to change { Sidekiq.schedule }.to(options[:schedule])
+      it { expect { subject }.to change { Sidekiq::Scheduler.enabled }.to(options[:enabled]) }
+      it { expect { subject }.to change { Sidekiq::Scheduler.dynamic }.to(options[:dynamic]) }
+      it { expect { subject }.to change { Sidekiq::Scheduler.listened_queues_only }.to(options[:listened_queues_only]) }
+      it { expect { subject }.to change { Sidekiq.schedule }.to(options[:schedule]) }
```
snmgian: Maybe this reads better ...
``` ruby
it { 
  expect { subject }.to change { Sidekiq.schedule }.to(options[:schedule]) 
}
```

spec/sidekiq-scheduler/schedule_spec.rb
```diff
@@ -1,85 +1,60 @@
...
+      subject
 
-      expect(cron_hash).to eq(job_from_redis(job_id))
-    end
+      job = Store.job_from_redis(job_id)
 
-    it 'uses job name as \'class\' argument if it\'s missing' do
-      Sidekiq.schedule = {job_class_id => cron_hash.select(&only_cron_and_args)}
+      expect(job).to eq(schedule)
+      expect(job['class']).to eq(schedule['class'])
```
snmgian: If the previous expect passes, this will pass also. Remove it

spec/sidekiq-scheduler/schedule_spec.rb
```diff
@@ -1,85 +1,60 @@
...
 
-      context 'when schedule is set twice' do
-        it 'sets the schedule on redis' do
-          2.times { Sidekiq.schedule = { job_id => cron_hash } }
+        job = Store.job_from_redis(job_id)
 
-          expect(cron_hash).to eq(job_from_redis(job_id))
-        end
+        expect(job).to eq(schedule.merge('class' => job_id))
+        expect(job['class']).to eq(job_id)
```
snmgian: idem

spec/sidekiq-scheduler/schedule_spec.rb
```diff
@@ -1,85 +1,60 @@
...
+      job = Store.job_from_redis(job_id)
 
-    it 'uses job name as \'class\' argument if it\'s missing' do
-      Sidekiq.schedule = {job_class_id => cron_hash.select(&only_cron_and_args)}
+      expect(job).to eq(schedule)
+      expect(job['class']).to eq(schedule['class'])
+    end
 
-      job = Sidekiq.schedule[job_class_id]
+    context "when 'class' argument is not set" do
```
snmgian: `'when "class" argument is not set'`

spec/support/store.rb
```diff
@@ -0,0 +1,16 @@
+class Store
+  class << self
```
snmgian: use `def self.method` instead of `class << self`

spec/support/store.rb
```diff
@@ -0,0 +1,16 @@
+class Store
```
snmgian: `module` instead of `class`

`Store` is a really generic name, some other gem could export a class/module with the same name and in that case we'll face some problems here

snmgian: scope it under SidekiqScheduler

spec/sidekiq-scheduler/job_presenter_spec.rb
```diff
@@ -1,116 +1,99 @@
...
 
-  describe "#next_time" do
-    context "when the job has not a next time in redis" do
-      it "returns nothing" do
-        expect(subject.next_time).not_to be
-      end
+  describe '#next_time' do
+    subject { job_presenter.next_time }
+
+    context "when the job doesn't have a next time in redis" do
```
mconiglio: So you are saying that there shouldn't be a context here?

snmgian: That's an option. Another option is to set the appropriate context using a `before` block

PR #201, opened 2017-07-13
Relax the rufus-scheduler dependency

nikolai-b:
I would like to use a new version (3.3.3) of rufus-scheduler which takes the time zone from Rails

https://github.com/jmettraux/rufus-scheduler/commit/dc9e9df051b2d9f5379f3cd5c703f078b35b8e3a

https://github.com/jmettraux/rufus-scheduler/blob/master/CHANGELOG.txt

This might mean some jobs are missed due to DST changes (i.e. if the before the TZ was UTC and now its Europe/London) so maybe this change should come with a warning...?

### DIFF
--- a/sidekiq-scheduler.gemspec
+++ b/sidekiq-scheduler.gemspec
@@ -16,7 +16,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'sidekiq',         '>= 3'
   s.add_dependency 'redis',           '~> 3'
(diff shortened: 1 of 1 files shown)


snmgian:
Hi @nikolai-b 

`~> 3.2` means `>= 3.2 && < 4.0`, so the PR will not have any effect. 

See http://guides.rubygems.org/patterns/

Issue #203, opened 2017-07-29
Wrong queue name when config.active_job.queue_name_prefix is set.


aruprakshit:
When u have this setup

```ruby
config.active_job.queue_name_prefix = "muse"
```

Then the actual queue name is being created as `muse_muse_<queue_name>`. The source of bug is

this [line](https://github.com/moove-it/sidekiq-scheduler/blob/5953fdfa3c7574b9aeb89bed96a232456600de89/lib/sidekiq/scheduler.rb#L240) . From here we pass a prefixed queue name and then Rails adds the prefix again to the queue name from [here](https://github.com/rails/rails/blob/cfb1e4dfd8813d3d5c75a15a750b3c53eebdea65/activejob/lib/active_job/enqueuing.rb#L47)

I have created a [sample app](https://gitlab.com/aruprakshit/sidekiq-scheduler-sample). Look at logs

# Sidekiq log

```sh
2017-07-29T21:13:31.883Z 10842 TID-ov8xv042w INFO: Running in ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin16]
2017-07-29T21:13:31.883Z 10842 TID-ov8xv042w INFO: See LICENSE and the LGPL-3.0 for licensing details.
2017-07-29T21:13:31.883Z 10842 TID-ov8xv042w INFO: Upgrade to Sidekiq Pro for more features and support: http://sidekiq.org
2017-07-29T21:13:31.909Z 10842 TID-ov8xv042w INFO: Loading Schedule
2017-07-29T21:13:31.909Z 10842 TID-ov8xv042w INFO: Scheduling SampleJob {"every"=>"15s", "class"=>"SampleJob", "queue"=>"muse_milestones"}
2017-07-29T21:13:31.910Z 10842 TID-ov8xv042w INFO: Schedules Loaded
2017-07-29T21:13:31.910Z 10842 TID-ov8xv042w INFO: Starting processing, hit Ctrl-C to stop
2017-07-29T21:13:47.096Z 10842 TID-ov8ya3lg8 INFO: queueing SampleJob (SampleJob)
```

# development.log

```sh
[ActiveJob] Enqueued SampleJob (Job ID: 1bf988a5-f147-4ee9-b4f0-ffa2fc605786) to Sidekiq(muse_muse_milestones)
[...]

PR #208, opened 2017-08-05
Don't explicitly set queue name for ActiveJobs

snmgian:
Asking for an ActiveJob's queue gets back the queue name including a prefix in case is set with
`config.active_job.queue_name_prefix`, and then setting that queue name explicitly while invoking
the job will turn to a duplicate prefix.

Changed to only set the queue name when one is configured via sidekiq.yml


snmgian:
See issue #203

PR #213, opened 2017-08-22
Support :dynamic_every option (default 5s) for :dynamic

soundasleep:
In some situations it doesn't make sense for a queue, which is
updated very rarely, to continually parse the scheduled queue
for new changes.


snmgian:
Thanks!

PR #220, opened 2017-10-08
Fix typo in gemspec

rdunlop:
The description on rubygems says "queueinga jobs"...doubt that was the intention.


snmgian:
Thanks :+1:

PR #228, opened 2017-11-09
Move Scheduler class inside SidekiqScheduler module

mconiglio:
* Also move sidekiq extensions to separate files


snmgian:
Released. `v2.2.0`

PR #231, opened 2017-11-16
Describe non-normal conditios against `cron`, and `at` only-once run

snmgian:
Describe non-normal conditios under which `cron` and `at` jobs could be pushed more than once when
runnning on multiple hosts.


benjaminwood:
👏 nice, thanks @snmgian.

PR #234, opened 2017-12-14
Use each_value instead of each using only value

284km:
Use "each_value" instead of "each" using only value.


snmgian:
Thanks

snmgian:
Released. `v2.2.0`

PR #249, opened 2018-03-29
Provide a shutdown option to rufus-scheduler

soundasleep:
I've been resolving an issue where calling clear_schedule! repeatedly on production is leaving behind hundreds of threads, which are all waiting waiting here:

```
rufus-scheduler-3.4.2/lib/rufus/scheduler/jobs.rb:287:in `pop'
rufus-scheduler-3.4.2/lib/rufus/scheduler/jobs.rb:287:in `block (2 levels) in start_work_thread'
rufus-scheduler-3.4.2/lib/rufus/scheduler/jobs.rb:285:in `loop'
rufus-scheduler-3.4.2/lib/rufus/scheduler/jobs.rb:285:in `block in start_work_thread'
```

This is what it looks like in production, before (and after) this commit:

threads-1

By default, calling `rufus_scheduler.stop` just lets [all of the work threads continue running](https://github.com/jmettraux/rufus-scheduler/blob/440400c5695c74b005e9baedc6a1d56c28e4c354/lib/rufus/scheduler.rb#L114) and they can leak to their hearts content. 

This commit makes all of the existing work threads join the current thread before clearing the schedule, meaning that the threads no longer leak by default.

### DIFF
--- a/lib/sidekiq-scheduler/scheduler.rb
+++ b/lib/sidekiq-scheduler/scheduler.rb
@@ -201,8 +201,8 @@ module SidekiqScheduler
 
       # Stops old rufus scheduler and creates a new one.  Returns the new
       # rufus scheduler
-      def clear_schedule!
-        rufus_scheduler.stop
+      def clear_schedule!(opt: :wait)
+        rufus_scheduler.stop(opt)
         @rufus_scheduler = nil
         @@scheduled_jobs = {}
         rufus_scheduler


snmgian:
Hi @soundasleep,

Thanks for finding this. I have just a couple of questions here.

- Why receiving the optional `opt` parameter here? https://github.com/moove-it/sidekiq-scheduler/pull/249/files#diff-deea2d80052450959df71ac43036b254R204

Instead of:
```ruby
def clear_schedule!
  rufus_scheduler.stop(:wait)
  ...
```  

- Are you calling `clear_schedule` directly? 
- Where you calling it from? 
- Are you calling it inside a sidekiq job?

soundasleep:
I'm not calling `clear_schedule` directly, no - and I'm only including it as an optional parameter so that if others *are* calling it, they can change the behaviour back to what it was before...

I've essentially got a sidekiq scheduled worker that runs:

```
  def call
    Sidekiq.schedule = generated_schedule
    Sidekiq::Scheduler.reload_schedule!
  end
```

So I call clear_schedule! only via reload_schedule!, and yes it's being called from inside a Sidekiq job.

snmgian:
Hi @soundasleep 

We still support ruby 1.9, so keyword argument `:opt` will not work. I submitted a new PR that it's almost your solution, and also added some unit tests: https://github.com/moove-it/sidekiq-scheduler/pull/253/files

PR #252, opened 2018-05-02
Allow intervals lower than 1 second

dalpo:
I'm currently trying to do a simple scheduler of a half second, but I noticed that the current implementation of the gem does not consider intervals lower than 1 second due to a `time.to_i` conversion. With this PR I changed the conversion into float.

I hope this could be useful.

Thank you.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,3 @@
-source 'http://rubygems.org'
+source 'https://rubygems.org'
 
 gemspec
--- a/lib/sidekiq-scheduler/redis_manager.rb
+++ b/lib/sidekiq-scheduler/redis_manager.rb
@@ -131,7 +131,7 @@ module SidekiqScheduler
       job_key = pushed_job_key(job_name)
       registered, _ = Sidekiq.redis do |r|
         r.pipelined do
-          r.zadd(job_key, time.to_i, time.to_i)
+          r.zadd(job_key, time.to_f, time.to_f)
           r.expire(job_key, REGISTERED_JOBS_THRESHOLD_IN_SECONDS)
         end
       end
--- a/spec/sidekiq-scheduler/redis_manager_spec.rb
+++ b/spec/sidekiq-scheduler/redis_manager_spec.rb
@@ -291,7 +291,7 @@ describe SidekiqScheduler::RedisManager do
     it 'should store the job instance' do
       subject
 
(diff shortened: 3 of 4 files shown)


snmgian:
Hi @dalpo 

This could lead to duplicate jobs in a scenario of multiple nodes. 

See https://github.com/jmettraux/rufus-scheduler/blob/master/lib/rufus/scheduler.rb#L569

What's your use case?

dalpo:
Hi @snmgian, thanks for your reply!

So this is my scenario:

I need to poll some http json APIs to track some status changes and do some stuffs when this happen, very quickly. 
At the moment everything is running on top of two sidekiq processes and probably I'm having duplicated jobs which are handled by `sidekiq-unique-jobs`.

At the beginning of this project the scheduling for these polling jobs were every few minutes (or hours) and sidekiq-scheduler worked perfectly :)

But now I need to change the polling every few milliseconds... 
I understand that this isn't the perfect fit for sidekiq-scheduler, but this kind of quick fix worked for me, at least for now... 

Now I'm considering to move the scheduling logics into a dedicated rufur-scheduler process and removing `sidekiq-scheduler`. 
What do you suggest? 

Thanks

snmgian:
Hi,

Those JSON APIs are within your control? 

In that case you could go backwards. Instead of querying those APIs in X msec intervals, you could push a sidekiq job from those APIs when some status change occurs. Or push something to a Redis list, and in a separate process you could use `LPUSH` or `BRPOPLPUSH`. Redis has an extensive but nice API.

I'd prefer that option in order to avoid hitting the API a lot of times just to found that no status changes occurred.

If you turn to use `rufus-scheduler` directly, keep in mind that its default tick frequency is 300 ms. You should configure that in case you need a higher frequency (ie, more ticks per second).

Also, depending on your specifics, you wouldn't need `rufus-shceduler` at all. Maybe a `while `, plus some exit condition, plus a `sleep` will do the job. Personally, I will not add a new dependency if I can simply roll on my own. `rufus-scheduler` has some niceties that could be worth it or not depending on your needs.

I like your PR, but right now we can't risk pushing duplicates. Capping minimum interval resolution to 1 sec is not the best solution at all though, I have some ideas to fix the duplicates problem at its root, that are triggering some of the incoming work.

dalpo:
Thanks for all your suggestions @snmgian!
 
Sadly, the JSON APIs aren't under my control. We are speaking about small embedded devices. In the future I hope to refactor everything and use Redis or a small MQTT server in the middle.

I will close the PR for now. I hope to contribute to this useful gem in some other features. 

Thanks for all your support and time :)

PR #254, opened 2018-05-14
Add note about integration with ActiveJob

dwightwatson:
Noticed that when using sidekiq-scheduler with ActiveJob you don't need to manually require it and include `Sidekiq::Worker` in each job. By just extending from `ApplicationJob` it's all still wired up and good to go. I think it's good to note that this is supported, otherwise other Rails developers might be needlessly putting the `require` and `include` boilerplate in each of their jobs.


snmgian:
Thanks!

PR #267, opened 2018-10-12
Use list group instead of table for improve web UI

huacnlee:
Our case there have a lot of description for each jobs.

The old web UI table can't not display will, so I try to make a change.

### After

It also support responsive view:


marcelolx:
@huacnlee Could you rebase, please? I think we can add this change, it really looks better.

Issue #269, opened 2018-11-29
Clarify how intervals work when Sidekiq restarts


henrik:
I noticed that a job that runs `every: "24h"` never seems to actually run.

My theory is that Heroku restarts our app every day, and intervals are reset every time Sidekiq is restarted. Do I have that right?

Would using `interval` instead "survive"  a Sidekiq restart?

Or is my best bet to use `cron`?

I'm happy to contribute a clarification to the README if someone with more insight into the code can help clear this up.

henrik:
Thank you so much, appreciate it! I'll find time to make a PR to clarify these things in the docs.

Nice idea about the `first_in`. That also seems to be in keeping with the spirit of Sidekiq, making things resilient and reentrant.

I guess the worker-only dyno would still restart when code changes are pushed.

Heroku Scheduler would work, though I prefer to keep schedules in the repo.

I guess `cron` might or might not have issues depending. If the cron syntax effectively says "every 24 hours" it would have the same issue. If it says "at 01:23 am daily" I guess it would not be affected by restarts unless the scheduler happens to be down at that time.

snmgian:
The cron syntax is a bit rich, but it always translate to something like "at 01:23 am daily", it's not "every 24 hours counting from app start".

The idea of having a worker only dyno is to use it in combination with `cron`.

henrik:
Got it. I've now changed our "every" declarations to use the "cron" syntax instead – better to have them run at e.g. 1:23 every day and not be affected by restarts (unless very poorly timed). Thanks again so much for the explanations!

PR #270, opened 2018-12-06
README: Clarify how intervals work when Sidekiq restarts

henrik:
Based on the discussion in https://github.com/moove-it/sidekiq-scheduler/issues/269.


snmgian:
Thanks!

PR #277, opened 2019-04-17
Remove defunct reset method

jeffmeyers:
I was looking through the source here to find a way to reset my state without digging into Redis, and happened up the `reset` method - which, as of https://github.com/moove-it/sidekiq-scheduler/commit/70de80cbc6a1bc675bf4e741b8ce2169635059b1, points to a method that no longer exists.

I just removed it in case anyone else happened to be looking for similar functionality. Looks like only `start`/`stop` are leveraged in the Sidekiq startup hook.


marcelolx:
Thanks @jeffmeyers

PR #292, opened 2019-12-02
Add notes about enqueueing during restart/downtime to readme [ci skip]

### LINKED ISSUE #154: What happens while sidekiq is down
Hello,

When we are deploying, we stop sidekiq when we start the deployment process and restart if after. Sometime sidekiq is down for 2-3 minutes (assets compilation, new gems to install, ...). 
If we have a job that should be enqueued in this time frame, what happens ?
I was wondering if the job is enqueued anyway...

Thanks for your response !

Gil

Vasfed:
Fixes #154 (sort of)

This adds important information about job not being enqueued when sidekiq worker is down for any reason at job's scheduled time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,6 +312,19 @@ Non-normal conditions that could push a specific job multiple times are:
 
 `every`, `interval` and `in` jobs will be pushed once per host.
 
+## Notes on when sidekiq worker is down
+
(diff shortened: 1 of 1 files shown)


marcelolx:
Thanks @Vasfed

PR #297, opened 2019-12-28
Fix dependency for ruby 2.7.0

biow0lf:

PR #299, opened 2020-01-08
Ruby 2.7.0

PhilCoggins:
Same as https://github.com/moove-it/sidekiq-scheduler/pull/297, just locking in mock_redis lib, build should go green.


jaynetics:
@PhilCoggins 

note that `RUBY_VERSION` is evaluated at release-time and not at install-time, so you may want to leave it out. else this new part will be missing if you `rake release` with the version specified in `.ruby-version`.

rjherrera:
@PhilCoggins why is it necessary to include the `RUBY_VERSION >= '2.7.0'` condition? I know that those libraries are no longer bundled gems since 2.7.0 but in older ruby versions it would just be of no harm to include those `add_dependency` statements, as they are bundled gems by default, wouldn't it?

I also think that you accidentally included your local `.ruby-version` file.

Sorry to bother @snmgian but I didn't know who else to mention. I think this issue is very important since all projects updating to ruby 2.7.0 depending on this gem will be broken. Could you take a look to this PR?

rjherrera:
@snmgian thank you!

snmgian:
`v3.0.1` is released

Issue #305, opened 2020-06-15
Deprecation warning from redis-rb


jbwl:
I am now getting deprecation warnings in my sidekiq.log

`Redis#exists(key)` will return an Integer in redis-rb 4.3. 
`exists?` returns a boolean, you should use it instead. To opt-in to the new behavior now you can set Redis.exists_returns_integer =  true. To disable this message and keep the current (boolean) behaviour of 'exists' you can set `Redis.exists_returns_integer = false`, but this option will be removed in 5.0. 

```
(/var/www/bestappever/shared/bundle/ruby/2.6.0/gems/sidekiq-6.0.7/lib/sidekiq/launcher.rb:160:in `block (2 levels) in ❤')
```

doutatsu:
@jbwl This is Sidekiq issue, not scheduler one. And they have fixed it in the latest version, so you just need to bump the sidekiq gem

bpo:
The stacktrace in the issue description is from sidekiq, but I think the same issue is present in the scheduler codebase [here](https://github.com/moove-it/sidekiq-scheduler/blob/master/lib/sidekiq-scheduler/redis_manager.rb#L99)

doutatsu:
I see that makes sense then

jasonrudolph:
> This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs.

I'm pretty sure this issue should remain open. In related news, it looks like fix is underway in https://github.com/moove-it/sidekiq-scheduler/pull/306.

jasonrudolph:
> This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs.

Still seems like this issue should remain open. 😇

jasonrudolph:
> This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs.

I'm still seeing these warnings, so it still seems like this issue should remain open.

akhil-vineti:
This should remain open.

rmm5t:
Please review #329

marcelolx:
Fixed by https://github.com/moove-it/sidekiq-scheduler/pull/335

PR #306, opened 2020-07-23
Changed redis "exists" method calls to "exists?"

matt-may:
Redis "exists" method calls were changed to "exists?" to address a deprecation warning from the redis-rb gem.

Addresses https://github.com/moove-it/sidekiq-scheduler/issues/305.

### DIFF
--- a/lib/sidekiq-scheduler/redis_manager.rb
+++ b/lib/sidekiq-scheduler/redis_manager.rb
@@ -96,7 +96,7 @@ module SidekiqScheduler
     #
     # @return [Boolean] true if the schedules key is set, false otherwise
     def self.schedule_exist?
-      Sidekiq.redis { |r| r.exists(:schedules) }
+      Sidekiq.redis { |r| r.exists?(:schedules) }
     end
 
     # Returns all the schedule changes for a given time range.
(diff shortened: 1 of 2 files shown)


t27duck:
It's possible the `mock-redis` gem needs to be bumped for the tests to pass.

Also, since the sidekiq update requires the most recent version of the `redis` gem, the minimum version of sidekiq may need to be bumped in the gemspec as well.

iax7:
why this small change has delayed so much from merging?

akhil-gautam:
@matt-may Are there any plans to get this merged?

rmm5t:
> why this small change has delayed so much from merging?

Because the tests are failing.

> It's possible the mock-redis gem needs to be bumped for the tests to pass.

Agreed.

> Also, since the sidekiq update requires the most recent version of the redis gem which is driving this change, the minimum version of sidekiq may need to be bumped in the gemspec as well.

I think just the gemspec here needs to be updated.

```ruby
  s.add_dependency 'redis',           '~>= 4.2'
```

rmm5t:
See #329 which includes @matt-may's deprecation fix, but also fixes the specs as well while upgrading necessary dependencies.

marcelolx:
Thanks for the PR @matt-may! I'll be closing this since it has been fixed on https://github.com/moove-it/sidekiq-scheduler/pull/335

PR #317, opened 2020-11-13
Add support for ruby 2.7 to the 2.2 branch

boblail:


marcelolx:
@boblail Do you still need this? Is there something that is blocking you from upgrading sidekiq-scheduler to v3?

boblail:
@marcelolx, I don't think we need this anymore — we're able to upgrade to v3. Thanks!

PR #329, opened 2021-02-03
Fix deprecation warning for exists; Upgrade redis and mock_redis dependencies

### LINKED ISSUE #305: Deprecation warning from redis-rb

rmm5t:
## What does it do?

- Upgrade redis dependency
- Upgrade mock_redis
- Fix pending specs
- Use strings as keys instead of symbols

## What else do you need to know?

- The redis gem converts all keys from symbols to strings, but mock_redis does not. This PR helps remedy that discrepancy.

## Related Issues

- Fixes #305
- Closes #306


akhil-vineti (approved):
LGTM 🎉

akhil-vineti:
@snmgian Can we please get this merged?

jdongelmans:
Any update? It's been quite a while since these warnings started popping up and the list of PR's regarding this keeps growing 😉

etiennemartin:
We are also seeing these warnings in our build. 🙏 Can this be merged?

marcelolx:
Thanks for this @rmm5t! The deprecation warning already has been fixed on this PR https://github.com/moove-it/sidekiq-scheduler/pull/335.

I think we can consider merge the part related to using the keys as strings, but we don't want to drop support for ruby 2.2/2.3 and redis < 4.2, that would require a major release which is not the focus at the moment.

Feel free to send a PR with the changes related to the symbol keys, I'll be happy to review it!

PR #259, opened 2018-06-20
Sort jobs by name on the web UI

attilahorvath:
Sort schedule jobs by name. Closes https://github.com/moove-it/sidekiq-scheduler/issues/258.

PR #330, opened 2021-02-26
Sort recurring jobs by name in web view

c-sonnier:
When you have a lot of recurring jobs it makes it much easier to find the one you are looking for.


marcelolx:
@c-sonnier I think that is what happens already, this has been done by this PR https://github.com/moove-it/sidekiq-scheduler/pull/259 and released on v.3.0.1 https://github.com/moove-it/sidekiq-scheduler/releases/tag/v3.0.1

PR #316, opened 2020-10-28
Replace deprecated method call `URI.escape`

coderberry:
Replace deprecated method call `URI.escape` with the recommended `URI.encode_www_form_component` method

PR #331, opened 2021-02-26
Replace URI.escape with CGI.escape for Ruby 3 compatibility

ehutzelman:


ehutzelman:
Duplicate of https://github.com/moove-it/sidekiq-scheduler/pull/316

Issue #333, opened 2021-03-13
NoMethodError (undefined method `escape' for URI:Module)


stefanahman:
**Background**:

sidekiq: 5.2.1
sidekiq-scheduler: 3.0.1
ruby: 3.0.0

Gettings `NoMethodError (undefined method `escape' for URI:Module)` when visiting http://localhost:3000/sidekiq/recurring-jobs

**Issue**:

`URI.escape` and `URI.unescape` are deprecated and removed (https://github.com/ruby/uri/commit/61c6a47ebf1f2726b60a2bbd70964d64e14b1f98). The recommendations is to use `CGI.escape` instead (https://github.com/ruby/uri/issues/14#issuecomment-751342417).

PR #334, opened 2021-03-13
Add support for Ruby 3.0

stefanahman:
Use ERB::Util.url_encode instead of deprecated URI.escape (#333)
Add more rubies to test matrix

### DIFF
--- a/web/views/recurring_jobs.erb
+++ b/web/views/recurring_jobs.erb
@@ -34,10 +34,10 @@
             </span>
           </td>
           <td class="text-center">
-            <a class="btn btn-warn btn-xs" href="<%= root_path %>recurring-jobs/<%= URI.escape(job.name) %>/enqueue">
+            <a class="btn btn-warn btn-xs" href="<%= root_path %>recurring-jobs/<%= ERB::Util.url_encode(job.name) %>/enqueue">
               <%= t('enqueue_now') %>
             </a>
-            <a class="btn <%= job.enabled? ? "btn-primary" : "btn-warn"%> btn-xs" href="<%= root_path %>recurring-jobs/<%= URI.escape(job.name) %>/toggle">
+            <a class="btn <%= job.enabled? ? "btn-primary" : "btn-warn"%> btn-xs" href="<%= root_path %>recurring-jobs/<%= ERB::Util.url_encode(job.name) %>/toggle">
               <%= job.enabled? ? t('disable') : t('enable') %>
             </a>
           </td>
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: ruby
 rvm:
+  - 3.0
   - 2.7
   - 2.6
   - 2.5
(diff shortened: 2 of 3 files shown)


melvrickgoh (approved):
LGTM 💯

rposborne:
Anything we can do to help get this merged?

Grandvizir:
Hi @melvrickgoh, 

Would love if you can merge this PR or telling us what blocks you to do it so we can give you help :) 
Thanks a lot!

halorrr:
@snmgian is there anything this needs to be merged? Thanks!

haffla:
I think this is a duplicate of https://github.com/moove-it/sidekiq-scheduler/pull/316.

fabiodallazen:
@stefanahman,
Hi, what is the prediction of the merge of this correction?

stefanahman:
> @stefanahman,
> Hi, what is the prediction of the merge of this correction?

🤷🏻‍♂️ idk, who's maintaining this?

bcavileer:
@snmgian Would you please review this and deploy an update?

unrooty:
@gian-zas or @snmgian can you please  review this and deploy an update?

web/views/recurring_jobs.erb
```diff
@@ -34,10 +34,10 @@
             </span>
           </td>
           <td class="text-center">
-            <a class="btn btn-warn btn-xs" href="<%= root_path %>recurring-jobs/<%= URI.escape(job.name) %>/enqueue">
+            <a class="btn btn-warn btn-xs" href="<%= root_path %>recurring-jobs/<%= CGI.escape(job.name) %>/enqueue">
```
marcelolx, line 37: @stefanahman What do you think about using `ERB::Util.url_encode` instead of `CGI.escape`? As far as I know CGI.escape makes more sense to use when do you need escape query params, while `ERB::Util.url_encode` when you need encode a URL 🤔 

This escape isn't required because modern browsers already handle this, they switch the spaces with `%20`, I made a test removing the escape and as you can see below the browser replaced the spaces with `%20`
image

But since this was added to make sure the url would be encoded correctly, I think we can keep this, but instead of using `CGI.escape` use `ERB::Util.url_encode` to keep the same behavior that it has before and the same behavior the browsers have by default
```irb
irb(main):002:0> URI.escape 'Foo Job'
=> "Foo%20Job"
irb(main):003:0> CGI.escape 'Foo Job'
=> "Foo+Job"
irb(main):004:0> ERB::Util.url_encode 'Foo Job'
=> "Foo%20Job"
```

stefanahman: Absolutely! This makes more sense. Fixed in https://github.com/moove-it/sidekiq-scheduler/pull/334/commits/6d697153119d6e322d3568fbebf419aed4c07c75

marcelolx (approved):
LGTM

elpic (approved):
LGTM

marcelolx:
Thanks @stefanahman 🎉

haffla:
@marcelolx now that Ruby 3 is supported doesn't this warrant a new release, too?

marcelolx:
> @marcelolx now that Ruby 3 is supported doesn't this warrant a new release, too?

@haffla Yes, we're just running sidekiq-scheduler with master branch in our applications to ensure nothing has been broken. Next week probably we will ship a new release.

marcelolx:
@haffla v3.1.0 released!

PR #335, opened 2021-03-19
fix redis exists deprecation warning

Wolfer:
With redis 4.2.5 I get deprecated text at startup

```
Redis#exists(key)` will return an Integer in redis-rb 4.3. `exists?` returns a boolean, you should use it instead. To opt-in to the new behavior now you can set Redis.exists_returns_integer =  true. To disable this message and keep the current (boolean) behaviour of 'exists' you can set `Redis.exists_returns_integer = false`, but this option will be removed in 5.0. (/usr/local/rvm/gems/ruby-2.6.3/gems/sidekiq-scheduler-3.0.1/lib/sidekiq-scheduler/redis_manager.rb:99:in `block in schedule_exist?')
```

Method `exists?` was added recently and most likely will not be available to everyone(https://github.com/redis/redis-rb/blame/master/lib/redis.rb#L601)

### DIFF
--- a/lib/sidekiq-scheduler/redis_manager.rb
+++ b/lib/sidekiq-scheduler/redis_manager.rb
@@ -96,7 +96,13 @@ module SidekiqScheduler
     #
     # @return [Boolean] true if the schedules key is set, false otherwise
     def self.schedule_exist?
-      Sidekiq.redis { |r| r.exists(:schedules) }
+      Sidekiq.redis do |r|
+        if r.respond_to?(:exists?)
+          r.exists?(:schedules)
+        else
+          !!r.exists(:schedules)
+        end
+      end
     end
 
     # Returns all the schedule changes for a given time range.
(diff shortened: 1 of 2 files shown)


weilandia:
We may want to also add the change here: https://github.com/moove-it/sidekiq-scheduler/blob/41990aafa80a0f458e988733189fbf4b07e0894a/spec/support/store.rb#L61

Wolfer:
> We may want to also add the change here:
> 
> https://github.com/moove-it/sidekiq-scheduler/blob/41990aafa80a0f458e988733189fbf4b07e0894a/spec/support/store.rb#L61

done

marcelolx:
@Wolfer Soon as possible I'll take a look on this, the tests are broken but this is already fixed here https://github.com/moove-it/sidekiq-scheduler/pull/339

marcelolx (approved):

elpic (approved):
LGTM

marcelolx:
Thanks @Wolfer

PR #336, opened 2021-04-06
Moves Redis calls to .exists to .exists? per deprecation warning

weilandia:
Warning:

```
`Redis#exists(key)` will return an Integer in redis-rb 4.3. `exists?` returns a boolean, you should use it instead. To opt-in to the new behavior now you can set Redis.exists_returns_integer =  true. To disable this message and keep the current (boolean) behaviour of 'exists' you can set `Redis.exists_returns_integer = false`, but this option will be removed in 5.0.
```


weilandia:
Already a pr

PR #339, opened 2021-05-12
Fix specs with Sidekiq >= 6.2

marcelolx:
Since https://github.com/mperham/sidekiq/pull/4804 sidekiq requires to be configured explicitly with a rack session.


elpic (approved):
LGTM

Issue #322, opened 2020-12-30
CSRF on enable/disable job


such:
Hello,

Since enabling/disabling a recurring job uses a GET request, it is subject to CSRF attacks.
It should use PUT.

PR #342, opened 2021-06-06
Use POST for write web requests

such:
To prevent CSRF

Closes https://github.com/moove-it/sidekiq-scheduler/issues/322

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -13,6 +13,8 @@ require 'timecop'
 # Load all support files
 Dir["#{File.dirname(__FILE__)}/support/**/*.rb"].each { |f| require f }
 
+$TESTING = true
+
 RSpec.configure do |config|
 
   config.after(:each) do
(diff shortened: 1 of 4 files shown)


spec/spec_helper.rb
```diff
@@ -13,6 +13,8 @@
 # Load all support files
 Dir["#{File.dirname(__FILE__)}/support/**/*.rb"].each { |f| require f }
 
+$TESTING = true
+
```
marcelolx, line 17: Is this required or did you forget it here?

such: It is required because ironically using POST makes tests fail because CSRF protection kicks in.
This is how sidekiq deals with the issue (https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/web.rb#L154)

marcelolx (approved):

elpic (approved):
LGTM

marcelolx:
Thanks @such

PR #344, opened 2021-06-25
Use list group instead of table for improve web UI

marcelolx:
Based on work done on this PR https://github.com/moove-it/sidekiq-scheduler/pull/267

I just added support for the dark mode too.

Closes #267

### DIFF
--- a/web/assets/stylesheets/recurring_jobs.css
+++ b/web/assets/stylesheets/recurring_jobs.css
@@ -1,0 +1,23 @@
+.recurring-jobs { border-top-left-radius: 4px; border-top-right-radius: 4px; }
+.recurring-jobs .title { margin-bottom: 5px; }
+.recurring-jobs .title .name { font-weight: bold;}
+.recurring-jobs .info,
+.recurring-jobs .description { margin-bottom: 5px; }
+.recurring-jobs .actions { margin-bottom: 5px; }
+.recurring-jobs .status,
+.recurring-jobs .description { font-size: 12px; }
+.recurring-jobs .enqueue { margin-bottom: 0.5rem }
+
+.list-group-item {
+  background-color: #222;
+  color: white;
+  border: 1px solid #555;
+}
+
+@media (prefers-color-scheme: dark) {
+  .list-group-item {
+    background-color: #222;
+    color: white;
+    border: 1px solid #555;
+  }
+}
--- a/web/views/recurring_jobs.erb
+++ b/web/views/recurring_jobs.erb
@@ -1,30 +1,6 @@
+<link href="<%= root_path %>stylesheets/recurring_jobs.css" media="screen" rel="stylesheet" type="text/css" />
+
 <h3><%= t('recurring_jobs') %></h3>
-
-<style type="text/css">
-  .recurring-jobs { border-top-left-radius: 4px; border-top-right-radius: 4px; }
-  .recurring-jobs .title { margin-bottom: 5px; }
-  .recurring-jobs .title .name { font-weight: bold;}
-  .recurring-jobs .info,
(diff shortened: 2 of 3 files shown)


web/views/recurring_jobs.erb
```diff
@@ -1,50 +1,74 @@
...
+  }
+
+  @media (prefers-color-scheme: dark) {
+    .list-group-item {
+      background-color: #222;
+      color: white;
+      border: 1px solid #555;
+    }
+  }
+</style>
```
juan-apa, line 27: is there a reason why this is embedded in the html file instead of a separate css?

web/views/recurring_jobs.erb
```diff
@@ -1,50 +1,74 @@
...
+        <div class="info">
+          <div class="row">
+            <div class="col-md-4 class"><%= job['class'] %></div>
+            <div class="col-md-4 interval text-left"><%= t('interval') %>: <%= job.interval %></div>
+            <div class="col-md-4 args"><%= t('arguments') %>: <%= job['args'] %></div>
+          </div>
+        </div>
+        <div class="status row">
+          <div class="col-md-4 actions">
+            <form action="<%= root_path %>recurring-jobs/<%= ERB::Util.url_encode(job.name) %>/enqueue" method="post" class="enqueue">
```
juan-apa, line 53: is there not already a helper for this route that you can use here, where you pass the job's name via param?

juan-apa (commented):
code looks good, I'll get the approve when I finish testing this locally in different screen resolutions!

web/views/recurring_jobs.erb
marcelolx: Not that I'm aware, at least sidekiq doesn't uses something similar https://github.com/mperham/sidekiq/blob/master/web/views/scheduled.erb#L15

web/views/recurring_jobs.erb
marcelolx: Not a specific reason, we can move it to a css file without problems

juan-apa: because there is nowebpacker to precompile the scss/css I don't see keeping it in the html file would be bad. However because this file slimmer, I think keeping styles in a sepparate css file would be the best option. Is there a way to insert the `<link rel="stylesheet" href="styles.css">` into the head from here using `content_for(:head)`?

juan-apa: what are your thoughts on this?

marcelolx: Yes I think that would be a better option, tomorrow I'll take a look on how sidekiq does it push some changes

marcelolx: @juan-apa Using `content_for` won't work because sidekiq and sidekiq-scheduler are framework agnostic, and `content_for(:head)` is an ActionView functionality. 

I've moved the css to an specific file and added Rack::Static as a middleware to serve this file

juan-apa (approved):
great pr!

web/views/recurring_jobs.erb
juan-apa: yep, you are right, got so used to some of rails helpers that I assumed they were always there 😅.

web/assets/stylesheets/recurring_jobs.css
```diff
@@ -0,0 +1,23 @@
...
+  border: 1px solid #555;
+}
+
+@media (prefers-color-scheme: dark) {
+  .list-group-item {
+    background-color: #222;
+    color: white;
+    border: 1px solid #555;
+  }
+}
```
juan-apa, line 23: add a new line here

juan-apa (approved):

Issue #345, opened 2021-07-16
Exception `ERR unknown command 'exists?'` in 3.1.0


barrucadu:
Hello, I think the change introduced in #335 isn't quite right, as in my app the `r.exists?` case is always getting called, even though it's not supported by redis gem I'm using:

```ruby
irb(main):002:0> Sidekiq.redis { |r| r.exists?(:schedules) }
Traceback (most recent call last):
        3: from (irb):1
        2: from (irb):2:in `rescue in irb_binding'
        1: from (irb):2:in `block in irb_binding'
Redis::CommandError (ERR unknown command 'exists?')

irb(main):003:0> Sidekiq.redis { |r| !!r.exists(:schedules) }
=> false

irb(main):004:0> Sidekiq.redis { |r| r.respond_to?(:exists?) }
=> true
```

This is making the worker crash on start-up:

```
ERR unknown command 'exists?'     
/data/vhost/account-api/shared/bundle/ruby/2.7.0/gems/redis-4.1.4/lib/redis/client.rb:126:in `call'
/data/vhost/account-api/shared/bundle/ruby/2.7.0/gems/redis-4.1.4/lib/redis.rb:3312:in `block in method_missing'
/data/vhost/account-api/shared/bundle/ruby/2.7.0/gems/redis-4.1.4/lib/redis.rb:51:in `block in synchronize'
/usr/lib/rbenv/versions/2.7.2/lib/ruby/2.7.0/monitor.rb:202:in `synchronize'
/usr/lib/rbenv/versions/2.7.2/lib/ruby/2.7.0/monitor.rb:202:in `mon_synchronize'
/data/vhost/account-api/shared/bundle/ruby/2.7.0/gems/redis-4.1.4/lib/redis.rb:51:in `synchronize'
/data/vhost/account-api/shared/bundle/ruby/2.7.0/gems/redis-4.1.4/lib/redis.rb:3311:in `method_missing'
/data/vhost/account-api/shared/bundle/ruby/2.7.0/gems/redis-namespace-1.8.1/lib/redis/namespace.rb:476:in `call_with_namespace'
/data/vhost/account-api/shared/bundle/ruby/2.7.0/gems/redis-namespace-1.8.1/lib/redis/namespace.rb:352:in `block (2 levels) in <class:Namespace>'
[...]

PR #346, opened 2021-07-17
Revert "fix redis exists deprecation warning"

marcelolx:
Reverts moove-it/sidekiq-scheduler#335

Ref: https://github.com/moove-it/sidekiq-scheduler/issues/345

### DIFF
--- a/lib/sidekiq-scheduler/redis_manager.rb
+++ b/lib/sidekiq-scheduler/redis_manager.rb
@@ -96,13 +96,7 @@ module SidekiqScheduler
     #
     # @return [Boolean] true if the schedules key is set, false otherwise
     def self.schedule_exist?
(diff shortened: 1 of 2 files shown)


marcelolx:
@juan-apa Just to give you more context, I think the best option to fix https://github.com/moove-it/sidekiq-scheduler/issues/345 is if we check the `redis` version and then chose which method call, but `mock_redis` only implements `exists?` on v0.24 https://github.com/sds/mock_redis/releases/tag/v0.24.0 so we will need to upgrade since we are using v0.19 and between this bumps support for some ruby versions was dropped which I think we should drop too since ruby 2.4 already reached end of life a year ago.

juan-apa (approved):

PR #347, opened 2021-07-17
Add support for redis-rb >= 4.3

marcelolx:
Follow-up of https://github.com/moove-it/sidekiq-scheduler/pull/346

When that PR gets merged we need to add support for redis-rb >= 4.3 which will return an integer value for `exists` while prior versions return a boolean value.

Fixing the warning described in original PR https://github.com/moove-it/sidekiq-scheduler/pull/335 will be made in a future PR and will be included only in sidekiq-scheduler v4 as described here https://github.com/moove-it/sidekiq-scheduler/issues/345#issuecomment-881971975

### DIFF
--- a/spec/support/store.rb
+++ b/spec/support/store.rb
@@ -59,7 +59,14 @@ module SidekiqScheduler
     end
 
     def self.exists(key)
-      Sidekiq.redis { |r| r.exists(key) }
+      Sidekiq.redis do |r|
+        case r.exists(key)
+        when true, 1
+          true
+        else
+          false
+        end
+      end
     end
 
     def self.hexists(hash_key, field_key)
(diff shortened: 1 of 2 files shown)


spec/support/store.rb
```diff
@@ -60,10 +60,11 @@ def self.zrange(zset_key, from, to)
...
       Sidekiq.redis do |r|
-        if r.respond_to?(:exists?)
-          r.exists?(key)
+        case r.exists(key)
+        when true, 1
+          true
         else
-          !!r.exists(key)
+          false
         end
```
juan-apa, line 68: 1 and true evaluate to a truthy value, so doing `r.exists(key) ? true : false` would do the trick here

marcelolx: @juan-apa Yeah, but `0` also will evaluate to a truthy value
```ruby
irb(main):001:0> true ? 'A' : 'B'
=> "A"
irb(main):002:0> 1 ? 'A' : 'B'
=> "A"
irb(main):003:0> 0 ? 'A' : 'B'
=> "A"
irb(main):004:0> false ? 'A' : 'B'
=> "B"
```

juan-apa: oh wow. I mean, makes sense, but at the same time it doesn't. It's fine by me if you keep it with the case then.

juan-apa (approved):

Issue #352, opened 2021-09-06
CSS file not loaded in scheduler UI


t27duck:
Commit 07b9380afc8e037789fcb02c3f7db9a8c779a0dd moved the styles to a separate file and attempted to mount it with the rest of the sidekiq assets.

It looks like the CSS file does not have its path properly set up.

(For good measure, I disabled cache in the Chrome network tab)

<img width="1455" alt="Screen Shot 2021-09-06 at 11 57 44 AM" src="https://user-images.githubusercontent.com/545604/132242301-587763de-f4fb-4107-a86c-88c0138bc43d.png">

PR #353, opened 2021-09-06
Fix CSS file path in web UI.

t27duck:
Commit 07b9380afc8e037789fcb02c3f7db9a8c779a0dd moved the styles to 
a separate file and mount like the rest of the Sidekiq assets.

It looks like the CSS file does not have its path properly set up.

The CSS path has been adjusted so that it loads properly now.

Before:

Screen Shot 2021-09-06 at 11 57 44 AM

After:

Screen Shot 2021-09-06 at 11 58 14 AM

Fixes: #352


marcelolx (approved):
Thanks @t27duck!

marcelolx:
@ujh We did change the layout, please check https://github.com/moove-it/sidekiq-scheduler/pull/344

PR #355, opened 2021-09-30
Fix CSS for light-mode users

t27duck:
The CSS that would run for light-mode users is identical to dark mode's styles.

Screen Shot 2021-09-30 at 7 49 22 AM

This adjusts the default styles to at least match closer to the rest of Sidekiq's UI for the default theme (values "borrowed" from Sidekiq's CSS file).

Screen Shot 2021-09-30 at 7 49 09 AM


marcelolx (approved):
Thanks!

PR #362, opened 2022-02-15
Add docs for running multi-sidekiq configurations

### LINKED ISSUE #361: Web: Running a 2nd sidekiq process will overwrite schedule shown in Recurring Jobs tab
I've noticed that, if one runs a 2nd sidekiq process with a separate configuration, the recurring jobs tab gets emptied out in the web extension.

**Please note** that the schedule itself with rufus still continues to operate and queue the scheduled jobs on the correct intervals.

I believe this might be the cause of #293 

## Setup
Run a rails app with Sidekiq and `sidekiq-scheduler`. Mount the `Sidekiq::Web` UI with the `sidekiq-scheduler/web` required.

Include a `config/sidekiq.yml` with a configured schedule, e.g.,
```yaml
:concurrency: 5
:queues:
  - critical
  - default
  - low
:schedule:
  do_something_every_minute:
    class: DoSomethingJob
    args: matey
    queue: :scheduler
    cron: '0 * * * * * America/Los_Angeles'
# ...
```

Next create _another_ sidekiq configuration, this time **without** a schedule:
```yaml
:concurrency: 5
:queues:
  - scheduler
  - other
```

## Steps to reproduce

1. Start the rails server
2. Start a sidekiq process with the default configuration loaded (the one with the schedule)
3. Go to the Recurring Jobs tab in the Sidekiq Web UI, you should see the schedule like so:
image

4. Start the _other_ sidekiq process loading the configuration without a schedule
5. Refresh the Recurring Jobs tab, you'll see the schedule disappear:
image

This bug is order dependent, if the process with the schedule boots last, the Recurring Jobs tab will show the schedule.

I made a reproducible app here: https://github.com/wspurgin/vanilla/tree/sidekiq-scheduler

## Hypothesis

I believe it has to do with how the schedule is stored. Namely, that the [key used to store and retrieve](https://github.com/moove-it/sidekiq-scheduler/blob/bc2eaf3d69d04420490e132f068c0a29e99599da/lib/sidekiq-scheduler/redis_manager.rb#L42-L48) the schedules is just `schedules` and not namespaced to the booted Sidekiq process (e.g., by `pid` or full process name). Therefore the last booted sidkeiq process' "schedule" configuration wins out.

wspurgin:
Closes #361 with some additional docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,17 +334,13 @@ Possible solutions include:
 
 Each option has it's own pros and cons. 
 
-## Notes on when running multiple Sidekiq processors on the same Redis
+## Notes when running multiple Sidekiq processors on the same Redis
 
-This gem stores the configured schedule in Redis on boot. It's used, primarily,
-to display in the Web Integration, and allow you to interact with that schedule
-via that integration.
+### TL;DR
 
-If you're running multiple Sidekiq processes on the same Redis namespace with
-different configurations, **you'll want to explicitly _disable_ Sidekiq
-Scheduler** for the other processes not responsible for the schedule. If you
-don't, schedule that gets stored in Redis will be determined by the last booted
-sidekiq process.
+Be **sure** to include the `:enabled: false` top-level key on any additional
+configurations to avoid any possibility of the `schedules` definition being
+wiped by the second Sidekiq process.
 
 To illustrate what we mean:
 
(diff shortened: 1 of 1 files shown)


marcelolx:
Thank you! ❤️

Issue #368, opened 2022-03-30
Highlight disabled jobs


duderamos:
Currently the only clue to know if a job is enabled or disabled is the toggle button. In my case here, that can have 10+ jobs and different ones enabled/disabled depending on the environment, this becomes uneasy to see.

So, what if we had a different background color in on the disabled?
<img width="1200" alt="Screen Shot 2022-03-30 at 15 03 45" src="https://user-images.githubusercontent.com/1532512/160755500-86664040-f18e-402a-8abc-2da81b78b62e.png">

PR #369, opened 2022-03-30
Highlight disabled jobs

duderamos:
This change adds a new CSS class to make disabled jobs more easily identifiable.

This PR resolves this issue: https://github.com/moove-it/sidekiq-scheduler/issues/368

### DIFF
--- a/web/assets/stylesheets/recurring_jobs.css
+++ b/web/assets/stylesheets/recurring_jobs.css
@@ -14,6 +14,10 @@
   border: 1px solid rgba(0, 0, 0, 0.1);
 }
 
+.list-group-item-disabled {
+  background-color: #f3d3d3;
+}
+
 @media (prefers-color-scheme: dark) {
   .list-group-item {
     background-color: #222;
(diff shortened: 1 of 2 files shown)


marcelolx:
Hey @duderamos, thanks! Have you tested how it does look with the dark mode? Based on the color I think it probably will be ok, but if you would have a screenshot of how it does look, it would be helpful.

marcelolx:
Also, I have cherry-picked your change to the [v3-stable](https://github.com/moove-it/sidekiq-scheduler/tree/v3-stable) branch, please point your Gemfile to that branch until a new release is made.

duderamos:
That's awesome, thanks @marcelolx . I will copy the screenshot from the issue into this PR description.

marcelolx:
@duderamos I did see that screenshot from the issue, what I would like to see would be one to see how it would look like with the dark mode (dark mode is enabled if dark mode is enabled on your OS System)... But that's not necessarily needed, if it may cause any issue someone will come up with a PR or open an issue, and then we can address it (very unlikely, but not impossible).

Thank you again!

Issue #374, opened 2022-04-07
Custom rufus_scheduler_options does not take into effect during runtime when dynamic: true


shayonj:
## Problem

When you set a custom option like the following for rufus scheduler, it does not take into effect during runtime.

```
SidekiqScheduler::Scheduler.instance.rufus_scheduler_options = { max_work_threads: 5 }
```
Readme: https://github.com/moove-it/sidekiq-scheduler#notes-about-connection-pooling

## How / Why

In a rails like application, this is primarily since gems are initialized first. So `sidekiq-scheduler`'s sidekiq `startup` hook gets defined first - https://github.com/moove-it/sidekiq-scheduler/blob/41990aafa80a0f458e988733189fbf4b07e0894a/lib/sidekiq-scheduler.rb#L12

The startup hook/event calls the `start` function https://github.com/moove-it/sidekiq-scheduler/blob/41990aafa80a0f458e988733189fbf4b07e0894a/lib/sidekiq-scheduler.rb#L18

Which ends up calling the `load_schedule!` function and ultimately the memoized `rufus_scheduler` instance variable [here](https://github.com/moove-it/sidekiq-scheduler/blob/51fc23863317e11fde57678be985398b0aa2f8be/lib/sidekiq-scheduler/scheduler.rb#L76) when `dynamic` is set to `true` in the config options.

Since during application boot time `@rufus_scheduler` is already instantiated, setting a custom `SidekiqScheduler::Scheduler.instance.rufus_scheduler_options` ([getter](https://github.com/moove-it/sidekiq-scheduler/blob/51fc23863317e11fde57678be985398b0aa2f8be/lib/sidekiq-scheduler/scheduler.rb#L170)/[setter](https://github.com/moove-it/sidekiq-scheduler/blob/51fc23863317e11fde57678be985398b0aa2f8be/lib/sidekiq-scheduler/scheduler.rb#L174)https://github.com/moove-it/sidekiq-scheduler/blob/51fc23863317e11fde57678be985398b0aa2f8be/lib/sidekiq-scheduler/scheduler.rb#L76) from the application's initializer or the application's own sidekiq startup hook does take into effect anymore.

[...]

PR #375, opened 2022-04-07
Set rufus_scheduler_options via sidekiq.yml file as configuration option

shayonj:
This ensures that `rufus_scheduler_options` is read during boot and
`sidekiq-scheduler` instance (that also instantiates rufus scheduler)
is initialized with any custom options passed into `rufus_scheduler_options`.

Doing it this way allows us to not set the `rufus_scheduler_options`, manually
during application runtime (like below), which doesn't take into effect, due to the
memoization of `@rufus_scheduler`.

```ruby
SidekiqScheduler::Scheduler.instance.rufus_scheduler_options = { max_work_threads: 5 }
```

More about the issue here: https://github.com/moove-it/sidekiq-scheduler/issues/374
Fixes: https://github.com/moove-it/sidekiq-scheduler/issues/374

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,7 @@ Available options are:
 :enabled: <enables scheduler if true [true by default]>
 :scheduler:
   :listened_queues_only: <push jobs whose queue is being listened by sidekiq [false by default]>
+:rufus_scheduler_options: <Set custom options for rufus scheduler, like max_work_threads [{} by default]>
 ```
 
 ## Schedule configuration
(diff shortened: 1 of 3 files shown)


marcelolx (approved):
LGTM! 

Since it is a breaking change it will only be released in v4. Until we release v4 you can point your Gemfile to the master branch

marcelolx:
I know it is fixing an issue but still is a breaking change so I prefer not to cherry-pick it to v3; v4 has no big changes except the minor version requirement for the redis gem so it should be a smooth upgrade path.

shayonj:
Sounds great! Thanks @marcelolx! Just to get an idea - by when do you think a v4 release is possible?

marcelolx:
Probably at the end of this month

PR #382, opened 2022-06-08
Remove Sidekiq::Util

mihaic195:
Fixes #381

PR #383, opened 2022-06-08
fix: add Sidekiq::Component

mihaic195:
Based on https://github.com/moove-it/sidekiq-scheduler/pull/382#issuecomment-1149775977 
Thanks for pointing that out. Indeed, it seems it was just replaced by a different name. @ltello @marcelolx

### DIFF
--- a/lib/sidekiq-scheduler/manager.rb
+++ b/lib/sidekiq-scheduler/manager.rb
@@ -1,5 +1,7 @@
 require 'redis'
 
+require 'sidekiq/component'
+
 require 'sidekiq-scheduler/schedule'
 require 'sidekiq-scheduler/scheduler'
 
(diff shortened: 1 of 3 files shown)


ltello:
Cool

marcelolx:
@ltello @mihaic195 Please see my comments on https://github.com/moove-it/sidekiq-scheduler/pull/382

mihaic195:
> @ltello @mihaic195 Please see my comments on #382

Good point, indeed. I agree.
You can close this PR if you think it's not relevant.

marcelolx:
I've also added one more comment https://github.com/moove-it/sidekiq-scheduler/pull/382#issuecomment-1149803452, basically, we aren't even using currently `sidekiq/util`, so one more reason to not depend on `Sidekiq::Component`.

Anyway, thank you both for your help! I really appreciate it ❤️

PR #385, opened 2022-06-10
Remove config get/set depication warning due to Sidekiq 6.5+

### LINKED ISSUE #384: `config.options[:key] =` is deprecated on Sidekiq 6.5.0
Hi,

I'm getting a deprecation message when using sidekiq-scheduler 4.0.1 with Sidekiq 6.5.0:

```
2022-06-08T14:47:01.226Z pid=46635 tid=trz WARN: `config.options[:key] = value` is deprecated, use `config[:key] = value`: ["sidekiq-scheduler-4.0.1/lib/sidekiq-scheduler.rb:16:in `block (2 levels) in <main>'", "sidekiq-6.5.0/lib/sidekiq/component.rb:56:in `block in fire_event'"]
```

Ref: mperham/sidekiq@67daa7a408b214d593100f782271ed108686c147

t27duck:
Closes #384 

Sidekiq 6.5 and later changed how config options work.

```ruby
 # Before
config.options[:x] = 1
config.options[:y] # => 2

 # After
config[:x] = 1
config[:y] # => 2
```

The existing functionality works, but now throws a deprication warning and will most likely be removed for Sidekiq 7. There is also no non-depricated way to access the config hash.

To maintain compatability with older Sidekiq versions, a `SIDEKIQ_GTE_6_5_0` has been added which is `true` if `Sidekiq::VERSION` is 6.5 or later. This constant is then used to determine now to pull the Sidekiq config hash and set configuration options where applicable.

This should allow the gem to maintain compatability with Sidekiq < 6.5

### DIFF
--- a/lib/sidekiq-scheduler.rb
+++ b/lib/sidekiq-scheduler.rb
@@ -7,19 +7,33 @@ require_relative 'sidekiq-scheduler/manager'
 require_relative 'sidekiq-scheduler/redis_manager'
 require_relative 'sidekiq-scheduler/extensions/schedule'
 
+SIDEKIQ_GTE_6_5_0 = Gem::Version.new(Sidekiq::VERSION) >= Gem::Version.new('6.5.0')
+
 Sidekiq.configure_server do |config|
 
   config.on(:startup) do
     # schedules_changed's type was changed from SET to ZSET, so we remove old versions at startup
     SidekiqScheduler::RedisManager.clean_schedules_changed
 
-    schedule_manager = SidekiqScheduler::Manager.new(config.options)
-    config.options[:schedule_manager] = schedule_manager
-    config.options[:schedule_manager].start
+    # Accessing the raw @config hash through .options is depriated in 6.5 and to be removed in 7.0
+    config_options = SIDEKIQ_GTE_6_5_0 ? Sidekiq.instance_variable_get(:@config) : config.options
+
+    schedule_manager = SidekiqScheduler::Manager.new(config_options)
+    if SIDEKIQ_GTE_6_5_0
+      config[:schedule_manager] = schedule_manager
+      config[:schedule_manager].start
+    else
+      config.options[:schedule_manager] = schedule_manager
+      config.options[:schedule_manager].start
+    end
   end
 
   config.on(:quiet) do
-    config.options[:schedule_manager].stop
+    if SIDEKIQ_GTE_6_5_0
+      config[:schedule_manager].stop
+    else
+      config.options[:schedule_manager].stop
+    end
   end
 
 end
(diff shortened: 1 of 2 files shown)


t27duck:
I have a very basic Rails app with sidekiq 6.5 and this branch wired up to enqueue an ActiveJob and stright Sidekiq job every 5 minutes.

https://github.com/t27duck/sidekiq_scheduler_384_test

- Pull
- Run `bundle install`
- `bundle exec sidekiq`
- `rails s` and http://localhost:3000/sidekiq should work including the schedules page

The jobs enqueue and run as expected. Deprication warnings no longer showed up.

🤷‍♂️

t27duck:
What I find odd is the specs are using the now depricated setters/getters for the Sidekiq config but I didn't see any deprication warnings printed out?

tagliala:
> What I find odd is the specs are using the now depricated setters/getters for the Sidekiq config but I didn't see any deprication warnings printed out?

I've tried to remove the code inside `config.on(:startup)` block and the specs keep passing

lib/sidekiq-scheduler.rb
```diff
@@ -7,19 +7,33 @@
...
 Sidekiq.configure_server do |config|
 
   config.on(:startup) do
     # schedules_changed's type was changed from SET to ZSET, so we remove old versions at startup
     SidekiqScheduler::RedisManager.clean_schedules_changed
 
-    schedule_manager = SidekiqScheduler::Manager.new(config.options)
-    config.options[:schedule_manager] = schedule_manager
-    config.options[:schedule_manager].start
+    # Accessing the raw @config hash through .options is depriated in 6.5 and to be removed in 7.0
```
olleolleolle, line 18:
```suggestion
    # Accessing the raw @config hash through .options is deprecated in 6.5 and to be removed in 7.0
```

olleolleolle:
Looks good! 👍

marcelolx:
> What I find odd is the specs are using the now depricated setters/getters for the Sidekiq config but I didn't see any deprication warnings printed out?

That's due to the log level in specs be in ERROR mode https://github.com/moove-it/sidekiq-scheduler/blob/master/spec/support/initialization/logs.rb#L5

t27duck:
`SidekiqScheduler::Manager` is assuming a hash-like object that can be merged into another hash.

Specifically, the whole thing breaks down here https://github.com/moove-it/sidekiq-scheduler/blob/master/lib/sidekiq-scheduler/manager.rb#L40

Sidekiq 6.5+'s new config object only exposes hash-like setters and getters for its keys, but there's no way to get the entire key-value set of options.

marcelolx:
@bpo @t27duck I think that for now to just fix the warning using `instance_variable_get(:@config)` is the best option we have ;/

We definitely want to pass the config object but that means a large refactoring to the manager which probably will be incompatible with older versions of sidekiq and will require a major bump.

marcelolx:
I want to make some more tests tomorrow, if everything looks good I'll push a new release, thanks @t27duck!

PR #388, opened 2022-06-21
Fix undefined method error

eldemcan:
Rufus used `jobs` method instead of `all_jobs`
https://github.com/jmettraux/rufus-scheduler/blob/3fe5a02b4ceae7901228010a0b58e62a3be0e456/lib/rufus/scheduler.rb#L277


marcelolx:
@eldemcan does this error happen with a specific version of Rufus?

eldemcan:
> @eldemcan does this error happen with a specific version of Rufus?

this error observed `sidekiq-scheduler-4.0.1`  `rufus-scheduler 3.8.1` 

image

PR #390, opened 2022-06-27
Add Dependabot for GitHub Actions

petergoldstein:
Adds Dependabot for GitHub Actions, ensuring that the action versions stay up to date.

Since the checkout action is currently out of date, if and when this is merged Dependabot will open a PR to update the action.


marcelolx:
thanks @petergoldstein!

PR #394, opened 2022-07-23
rename stylesheets, it conflict with sidekiq rake asset file

houdl:
stylesheets has been used for sidekiq . these will conflict with sidekiq asset file.
code is here: https://github.com/mperham/sidekiq/blob/main/lib/sidekiq/web.rb#L150

so we need use Rack::Static, urls: ["stylesheets-scheduler"]

image


marcelolx (approved):

marcelolx:
Thank you!

Issue #332, opened 2021-03-08
duplicate scheduling when running on multiple host


rvyas:
This gem works fine under general use case. However, we have some jobs that take longer time (20-25 min) and increases the queue latency.

During such scenario, jobs are scheduled twice which causes other issues in the app. There is a section in the readme that warns about this but is there a way i can minimize duplication? like tweaking certain parameters or limiting certain things?

OR can this issue be fixed in the gem itself?

Current config:
* 4 hosts
* 4 processes (1 per host)
* 12 threads each process
* 5 queue
* 6 recurring jobs using cron at different times.
* There can be >100 jobs scheduled in a queue at the same time.

bpo:
@rvyas you might want to take a look at [sidekiq-uniq](https://github.com/meedan/sidekiq-uniq)

rvyas:
Thanks @bpo. I will try this gem. 

I was trying to do the same in the worker/checking other jobs in queue, etc, but there is some delay when those jobs show up in sidekiq APIs.

bpo:
@rvyas I don't use that one myself, but I know others have had similar issues. Sidekiq Enterprise has [built-in support for unique jobs](https://github.com/mperham/sidekiq/wiki/Ent-Unique-Jobs), which would probably be your best choice if that's an option for you.

rvyas:
No, it did not solve the problem. Even with sidekiq-uniq I see the same duplicate scheduling although the rate has reduced.
May be I need to leverage built-in support.

mgoggin:
@ryvas Were you able to solve your issue with sidekiq-cron?

marcelolx:
@rvyas @mgoggin I didn't take a look at this gem https://github.com/mhenrixon/sidekiq-unique-jobs, but maybe it can help.

marcelolx:
ok I think one solution for those problems will be to use a Redis based Lock instead of rufus-scheduler default filelock

https://github.com/jmettraux/rufus-scheduler#advanced-lock-schemes

This would allow us to use something like https://github.com/kenn/redis-mutex to create a single lock to all schedulers running on different hosts which would ensure that only one scheduler would be able to schedule a job at a given time.

I'll investigate more and make some tests, this may land as an experimental feature in v4.1

bpo:
@marcelolx this seems like a tricky thing for sidekiq-scheduler to handle well, as it's still always possible for the app to enqueue jobs externally from the scheduler.

[...]

Issue #156, opened 2016-12-09
Double execution of scheduled jobs when multiple hosts are running


bbozo:
Three days ago we had a job run twice on 2 hosts after turning on Sidekiq on the 2nd host five days ago. We are using `sidekiq-scheduler (2.0.9)` and we set it up using this code in a non-Rails app:

``` ruby
Sidekiq.configure_server do |config|
  threads = config.options[:concurrency] = ( ENV['MAX_WORKER_THREADS'] || 4 ).to_i
  config.redis = ConnectionPool.new(size: threads + 1, &redis_conn)
  config.options[:verbose] = App.log.log_level == :debug

  # set up Sidekiq Scheduler
  config.on(:startup) do
    schedule = YAML.load_erb_file(App.root + "/config/scheduler.yml")
    Dir[App.root + "/modules/*/config/scheduler.yml"].each do |yaml|
      schedule.merge! YAML.load_erb_file(yaml)
    end

    Sidekiq.schedule = schedule
    Sidekiq::Scheduler.reload_schedule!
  end
end
```

This code has been working well for 6 months now before we added the 2nd host, the scheduler web interface shows that the jobs are correctly loaded.

If this helps, it's a very low latency environment, clocks synced perfectly, not an AWS/Heroku etc setup. However I always make sure that not all sidekiq's get restarted at the same time to avoid #154. I'm kind of afraid to try this again because the implications of a double run of a task are pretty terrifying in my use case.

I noticed today that a different user [makes sure not to run sidekiq scheduler on all machines](https://github.com/moove-it/sidekiq-scheduler/issues/147#issuecomment-255490978) yet there's nothing in the docs suggesting that this might be important.

bbozo:
I went through the code, it seems that this project simply uses Rufus to do the task running, but Rufus seemingly uses a file locking mechanism at the best and no locking at the worst so this would explain everything. I wrote to rufus people (that sounds wrong ^_^) to check out if there's any option to implement distributed locking in Rufus that could maybe be used in `sidekiq-scheduler`

But in the meantime, we need a big fat juicy warning for this and an example scheduling code that takes this into account.

bbozo:
[...]

PR #396, opened 2022-07-23
Before enqueueing the job register it with the exact time it was supposed to run

### LINKED ISSUE #181: Use rufus job's schedule time to determine if a specific job instance has ran before

marcelolx:
TL;DR; Work extracted from #276, thanks @afrase!

#156 had some interesting discussion around this and why it is more important to fix this first before trying to introduce locking which not necessarily would avoid double execution of scheduled jobs.

# Details

Rufus calls the block with the current time instead of the time the job is scheduled to run. If you have multiple hosts, it's possible
that Rufus will run the block at slightly different times which results in a job getting run multiple times.

This figures out when the job was scheduled to run instead of the time Rufus sends.

Co-Authored-By: Aaron Frase <afrase@users.noreply.github.com>

TODO:
- [ ] Tests
- [ ] Detail commit message/PR description
- [ ] Check if README needs to be updated

Closes #181
Closes #276

This may solve #332 too, but I need to think more about that scenario (at least A).

### DIFF
--- a/lib/sidekiq-scheduler/scheduler.rb
+++ b/lib/sidekiq-scheduler/scheduler.rb
@@ -234,7 +234,17 @@ module SidekiqScheduler
       options = options.merge({ :job => true, :tags => [name] })
 
       rufus_scheduler.send(interval_type, schedule, options) do |job, time|
-        idempotent_job_enqueue(name, time, SidekiqScheduler::Utils.sanitize_job_config(config)) if job_enabled?(name)
+        if job_enabled?(name)
+          conf = SidekiqScheduler::Utils.sanitize_job_config(config)
+
+          if job.is_a?(Rufus::Scheduler::CronJob)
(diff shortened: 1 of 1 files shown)


bhaskar-simpl:
@marcelolx Do you need help closing and releasing this?

marcelolx:
@bhaskar-simpl I have had no time lately to write some tests for this and test it, so any help is welcome; feel free to send a PR against the branch of this one or master.

marcelolx:
Closing in favor of #463

Issue #400, opened 2022-08-23
Allow for multiple applications to share a Redis DB


TALlama:
I have multiple applications that need to share a single Redis DB but have different scheduled jobs. The current codebase does not allow for this; the applications will all use the same keys and overwrite each other.

Currently, the Redis keys are static and unchangeable. Since they are referenced in a variety of locations (though some are centralized in `RedisManager`), a monkey patch to enable this kind of multi-app setup is prohibitively fiddly.

marcelolx:
Fixed in #401

PR #401, opened 2022-08-23
Allow for multiple applications to share a Redis DB

TALlama:
Implements #400

Adds the ability to set a `key_prefix` which will be prepended to all Redis keys. This allows multiple applications to keep their schedules in the same Redis database without clobbering each other.

This change does not affect the runtime of the RSpec suite, but allows for the use case mentioned. Doing the first two commits without the last one would allow for monkey patches, but if we could do the whole thing that would be great.

### DIFF
--- a/README.md
+++ b/README.md
@@ -418,6 +418,20 @@ Redis.
 
 See https://github.com/sidekiq-scheduler/sidekiq-scheduler/issues/361 for a more details.
 
+## Notes when running multiple applications in the same Redis database
+
+If you need to run multiple applications with differing schedules, the easiest way is to use a different Redis database per application. Doing that will ensure that each application will have its own schedule, web interface and statistics.
+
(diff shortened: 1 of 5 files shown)


marcelolx (commented):
Can we get some documentation about it added to the README.md?

TALlama:
@marcelolx added.

TALlama:
Also heads up that the link to [the commit message style guide](https://github.com/sidekiq-scheduler/guides/tree/master/style/commit-message#commit-message-style-guide) on [CONTRIBUTING.md](https://github.com/sidekiq-scheduler/sidekiq-scheduler/blob/master/CONTRIBUTING.md) is broken

marcelolx:
Thanks @TALlama! I'll look into updating the contributing guide.

Added the PR to my list to review, as soon as I have time I'll take a final look

marcelolx (approved):
Thank you @TALlama! It looks good to me, there is just a conflict, once that is fixed I'll merge this PR (if you don't have time, don't mind, I can fix it)

marcelolx:
Thanks @TALlama! I have to fix the CI to run the tests, but from what I remember the last time it did pass, so probably nothing is broken.

Thank you for your patience!

Issue #402, opened 2022-08-23
When `include_metadata: true`, `Sidekiq.strict_args!` fails with "Job arguments must be native JSON types"


craigmcn:
I'm getting a Sidekiq error message "Job arguments must be native JSON types" when I include the `include_metadata: true` option in my schedule. The specific error is:
"ArgumentError: Job arguments to Actions::System::DestroyOldUserSessions must be native JSON types, see https://github.com/mperham/sidekiq/wiki/Best-Practices.
To disable this error, remove `Sidekiq.strict_args!` from your initializer."

Without `Sidekiq.strict_args!`, I get a deprecation warning:
"WARN: Job arguments to Actions::System::DestroyOldUserSessions do not serialize to JSON safely. This will raise an error in
Sidekiq 7.0. See https://github.com/mperham/sidekiq/wiki/Best-Practices or raise an error today
by calling `Sidekiq.strict_args!` during Sidekiq initialization."

Removing `include_metadata: true` resolves the issue.

Ruby version: 2.7.5
Rails version: 7.0.3.1
Sidekiq version: 6.5.4
Sidekiq Scheduler version: 4.0.2

Sidekiq initializer:
```ruby
Sidekiq.configure_server do |config|
  # url format 'redis://redis.example.com:7372/0'
  config.redis = {url: ENV["REDIS_JOBS_URL"]}

  config.strict_args!
end
```

Sidekiq YAML:
```yml
production:
  :concurrency: 5
:queues:
  - actions
  - default
  - low
  - mailers
:schedule:
  revoke_old_user_session:
    description: "Revoke all user sessions that have not been accessed in the past 24 hrs"
    cron: "0 1 * * * America/Los_Angeles" # At 1 min after midnight everyday
    class: "Actions::System::RevokeOldUserSessions"
    include_metadata: true
    args: [
      { invoked_by: "system" }
    ]
  destroy_old_user_session:
    description: "Destroy user sessions that have been revoked over 1 year ago"
    cron: "0 3 * * * America/Los_Angeles" # At 3 min after midnight everyday
    class: "Actions::System::DestroyOldUserSessions"
    include_metadata: true
    args: [
      { invoked_by: "system" }
    ]
```

I believe I have a working PR that I will submit.

marcelolx:
Fixed on #403

PR #403, opened 2022-08-23
Fix metadata for Sidekiq `strict_args!`

craigmcn:
- update `scheduled_at` metadata to string key, hash rocket syntax
- update spec to expect string key, hash rocket syntax
- fixes issue #402

### DIFF
--- a/lib/sidekiq-scheduler/scheduler.rb
+++ b/lib/sidekiq-scheduler/scheduler.rb
@@ -158,7 +158,7 @@ module SidekiqScheduler
       config = prepare_arguments(job_config.dup)
 
       if config.delete('include_metadata')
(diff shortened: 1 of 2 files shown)


marcelolx (approved):
Thank you @craigmcn!

yeonhoyoon:
This could potentially be a breaking change for some applications, if `scheduled_at` was being accessed by symbol key.
It might have been a better idea to put this into a minor version, not patch version.

For people visiting this issue after updating the gem to 4.0.3:
`metadata[:scheduled_at]` should be changed to `metadata["scheduled_at"] `

marcelolx:
@yeonhoyoon could you give an example of why you're accessing the `scheduled_at`?

PR #404, opened 2022-09-06
Redis 5.0 compatibility

casperisfine:
Fix: https://github.com/redis/redis-rb/issues/1143

- Remove `redis_mock` as it was hiding several bugs
- Explicitly cast time types to string, as redis-rb 5.0 no longer does it implicitly

### DIFF
--- a/lib/sidekiq-scheduler/utils.rb
+++ b/lib/sidekiq-scheduler/utils.rb
@@ -112,8 +112,10 @@ module SidekiqScheduler
     def self.new_rufus_scheduler(options = {})
       Rufus::Scheduler.new(options).tap do |scheduler|
         scheduler.define_singleton_method(:on_post_trigger) do |job, triggered_time|
-          SidekiqScheduler::Utils.update_job_last_time(job.tags[0], triggered_time)
-          SidekiqScheduler::Utils.update_job_next_time(job.tags[0], job.next_time)
+          if (job_name = job.tags[0])
+            SidekiqScheduler::Utils.update_job_last_time(job_name, triggered_time)
+            SidekiqScheduler::Utils.update_job_next_time(job_name, job.next_time)
+          end
         end
       end
     end
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,17 @@ jobs:
     runs-on: ubuntu-latest
     strategy: 
       matrix: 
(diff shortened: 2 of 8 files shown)


lib/sidekiq-scheduler/utils.rb
```diff
@@ -112,8 +112,10 @@ def self.sanitize_job_config(config)
     def self.new_rufus_scheduler(options = {})
       Rufus::Scheduler.new(options).tap do |scheduler|
         scheduler.define_singleton_method(:on_post_trigger) do |job, triggered_time|
-          SidekiqScheduler::Utils.update_job_last_time(job.tags[0], triggered_time)
-          SidekiqScheduler::Utils.update_job_next_time(job.tags[0], job.next_time)
+          if (job_name = job.tags[0])
```
casperisfine, line 115: Not too sure why, but in some cases `job.tags == []`, which means we call `update_job_last_time(nil, triggered_time)`.

On redis-rb 4.x, `nil` would be silently casted to `""`, on 5.0 it raises (this is exactly the kind of errors I wanted to catch by making this change).

marcelolx: thanks @casperisfine, I'll take a look to see why in some cases `tags` is an empty array (added to my list of things to do 😅)

marcelolx (approved):
Thank you @casperisfine!

lib/sidekiq-scheduler/utils.rb
casperisfine: 👌 if you want to repro, it happens twice in the test suite, but inside the Rufus thread, so somehow the Rufus thread crash, but the test pass.

raivil:
@casperisfine are there any plans to release a new gem version with this fix?

casperisfine:
I'm not the maintainer.

raivil:
😅 sorry!
tagging the maintainer @marcelolx :)

marcelolx:
@raivil I plan to merge https://github.com/sidekiq-scheduler/sidekiq-scheduler/pull/396 before a release, so for now, please point your Gemfile to the master branch until a new release is made.

```ruby
gem "sidekiq-scheduler", github: "sidekiq-scheduler/sidekiq-scheduler", branch: "master"
```

marcelolx:
@raivil FYI v4.0.3 has been released

Issue #407, opened 2022-09-21
Scheduler uses Activesupport::Inflector#constantize without having dependency set in gemspec


swilgosz:
### Background

One of our teammates (@andrebalazs55) found that `sidekiq-scheduler` uses the `#constantize` method called on a string, which [comes from activesupport](https://apidock.com/rails/v6.1.3.1/ActiveSupport/Inflector/constantize). However, activesupport is not listed as a dependency in the gemspec file.

This causes issues in non-rails applications, which don't have activesupport in use.

### Possible Solutions

1. Add activesupport as a dependency (can have sideeffects across existing codebases)
2. Replace the `ActiveSupport::Inflector#constantize` with a small inflector gem (i.e: [dry-inflector](https://github.com/dry-rb/dry-inflector))
3. Replace the `ActiveSupport::Inflector#constantize` with the method definition.

PR #408, opened 2022-09-21
Fix the constantize method

swilgosz:
Resolves: #407

### Solution: 

Initially, I've thought about adding a [managed string inflector](https://github.com/dry-rb/dry-inflector/blob/main/lib/dry/inflector.rb#L88) as a dependency, but there is no need for that just to use this single method.

Instead, I've replaced the with the to avoid additional dependencies.


marcelolx (approved):
Thank you @swilgosz!

PR #411, opened 2022-11-02
Drop support for EOL Rubies & Sidekiq

marcelolx:
Based on the discussion we had here https://github.com/sidekiq-scheduler/sidekiq-scheduler/pull/410

I also fixed some broken specs for Ruby 3/3.1 that were expecting keyword arguments while they actually should expect a hash, which is what we push to sidekiq [here](https://github.com/sidekiq-scheduler/sidekiq-scheduler/blob/master/lib/sidekiq-scheduler/utils.rb#L86).


bpo (approved):
LGTM. I guess this will need to bump to a new major version?

t27duck (approved):

PR #412, opened 2022-11-04
Refactor: Use our own config class instead of mutating sidekiq config options hash

marcelolx:
Necessary changes to have our own config file and stop mutating Sidekiq config.

## Breaking changes

Scheduler options have to be under the `scheduler` key in the `sidekiq.yml` file, this includes the `schedule` key.

If any of the below keys is in the root of the `sidekiq.yml` file, we will raise an error telling you that the key should be under the `scheduler` key.

### before

```yml
:dynamic: false
:dynamic_every: false
:enabled: true
:scheduler:
  :listened_queues_only: false
:rufus_scheduler_options: {}
:schedule:
  HardJob:
    cron: '0 */5 * * * *' 
```

### after 

```yml
:scheduler:
  :dynamic: false
  :dynamic_every: false
  :enabled: true
  :listened_queues_only: false
  :rufus_scheduler_options: {}
  :schedule:
    HardJob:
      cron: '0 */5 * * * *' 
```


t27duck:
Took a swing at trying to get specs to pass for at least Sidekiq 7. I'm having issues understanding the manager specs and what they mean to do. Then hit a mental brick wall trying to figure out how to support Sidekiq 6.5 since Sidekiq introduced the config object in 7...

Just a thought: Since Sidekiq 7 is "the future" and had a good amount of under the hood changes, what are opinions on having this gem only support Sidekiq 7 going forward as the next major version bump? Part of the issue is `SidekiqScheduler::Config` is set up to want a sidekiq config object... I suppose we could have it pull the hash out of `Sidekiq` if it's version 6.5...

Alternatively, instead of dropping Sidekiq 6.x support, maybe make the bare minimum version of Sidekiq supported be 6.5?

t27duck:
... or not. 

Specs when I run Sidekiq 7 seem to all pass but the manager specs but I guess CI seems to have other issues... 😞

marcelolx:
thanks @t27duck! I did push some changes that fix most of the tests, and yeah, they only cover sidekiq 7 yet. 

The suite of tests failing with sidekiq 7 is due to a change in sidekiq, I sent a fix https://github.com/mperham/sidekiq/pull/5621

marcelolx:
I fixed the majority of the tests for both sidekiq versions, the only ones that are still broken are these here https://github.com/sidekiq-scheduler/sidekiq-scheduler/blob/master/spec/sidekiq-scheduler/manager_spec.rb#L110-L140, which in some way don't make much sense, but I have to debug the old logic to understand them. 

Regarding this ⬇️, we need to support both versions, it shouldn't be complicated, because in Sidekiq 6.x the config we receive in the `configure_server` block is a [Sidekiq module](https://github.com/mperham/sidekiq/blob/6-x/lib/sidekiq.rb#L142) while now it is the [Sidekiq::Config](https://github.com/mperham/sidekiq/blob/0e91f175fe8ada215bcf64bb086ca7585fe60e3f/lib/sidekiq.rb#L98) object, so we only need to pass one or another in our tests based on the Sidekiq version that is being tested against.

> Just a thought: Since Sidekiq 7 is "the future" and had a good amount of under the hood changes, what are opinions on having this gem only support Sidekiq 7 going forward as the next major version bump? Part of the issue is SidekiqScheduler::Config is set up to want a sidekiq config object... I suppose we could have it pull the hash out of Sidekiq if it's version 6.5...

t27duck:
I threw in a workaround and TODO until your PR for Sidekiq 7 gets merged and released to at least get the Sidekiq 7 specs going green and eliminate some noise. Easy enough to remove once that's sorted out.

lib/sidekiq-scheduler/sidekiq-adapter.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+
+    def self.start_schedule_manager(sidekiq_config:, schedule_manager:)
+      if SIDEKIQ_GTE_6_5_0
+        config[:schedule_manager] = schedule_manager
+        config[:schedule_manager].start
+      else
+        config.options[:schedule_manager] = schedule_manager
+        config.options[:schedule_manager].start
+      end
```
marcelolx, line 20:
```suggestion
      if SIDEKIQ_GTE_6_5_0
        sidekiq_config[:schedule_manager] = schedule_manager
        sidekiq_config[:schedule_manager].start
      else
        sidekiq_config.options[:schedule_manager] = schedule_manager
        sidekiq_config.options[:schedule_manager].start
      end
```
should this be ⬆️?

lib/sidekiq-scheduler/sidekiq-adapter.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+
+    def self.stop_schedule_manager
+      if SIDEKIQ_GTE_6_5_0
+        config[:schedule_manager].stop
+      else
+        config.options[:schedule_manager].stop
+      end
+    end
```
marcelolx, line 29: are we able to access the config object here somehow without receiving it as an argument of this method?

marcelolx (commented):
@t27duck I like this solution! I think the "Adapter" name makes sense for this use case 🎉

t27duck:
The joys of copy/pasting on code that isn't ran in tests. Thanks for pointing that out @marcelolx, should be fixed.

OskarEichler:
Any movement on this PR?

marcelolx:
Hopefully tomorrow I'll have some time to come up with a migration doc from v4 to v5, though it shouldn't be complicated since the only thing that really changes is that now everything will be under the `scheduler` key.

marcelolx:
A beta release is available for those who want to test and see if everything is working as expected! 

https://github.com/sidekiq-scheduler/sidekiq-scheduler/releases/tag/v5.0.0.beta1

PR #416, opened 2022-12-27
Drop explicit redis dependency

ursm:
Sidekiq 7.0 depends on redis-client by default, and earlier versions depend on redis.


marcelolx:
thanks @ursm!

PR #418, opened 2023-01-13
Round time to ms

### LINKED ISSUE #417: Sidekiq will periodically report JSON argument errors when using metadata on some machines
The code on [this line](https://github.com/sidekiq-scheduler/sidekiq-scheduler/blob/fa293cec9c95ccf6a4e4455b12071b30e664eb07/lib/sidekiq-scheduler/scheduler.rb#L173) will generate a floating point number for the `scheduled_at` time. On some machines this number has more precision than the JSON parser supports, as per the example below:

```
[4] pry(#<Sidekiq::Client>)> JSON.parse(JSON.dump(item["args"])) 
=> [{"scheduled_at"=>1673626980.005889}]
[5] pry(#<Sidekiq::Client>)> item["args"]
=> [{"scheduled_at"=>1673626980.0058892}]
```

This causes sidekiq to complain that the data passed in is not valid JSON, which may raise an exception in sidekiq 7 depending on the config. The simplest solution would be to change the `scheduled_at` value to an integer using `to_i` to get the resolution in seconds. If ms resolution is desired you can do `Time.new.to_f.round(3)` instead.

TikiTDO:
Fixes #417

### DIFF
--- a/lib/sidekiq-scheduler/scheduler.rb
+++ b/lib/sidekiq-scheduler/scheduler.rb
@@ -17,7 +17,7 @@ module SidekiqScheduler
     # saying we will have to do it somehow still)
     #
     # NOTE: ^ Keeping this TODO here for now, in a future version of this project
-    # we will remove those attr acessors and use only our config object. For now, 
+    # we will remove those attr acessors and use only our config object. For now,
     # let's keep as it is.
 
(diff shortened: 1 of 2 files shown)


marcelolx:
Thanks @TikiTDO!

PR #432, opened 2023-03-30
Fix "uppercase character in header name: Cache-Control"

### LINKED ISSUE #431: Uppercase character in header name: Cache-Control
Using [rack](https://github.com/rack/rack) > 3.0.0 requires all response header keys to be lowercase: https://github.com/rack/rack/commit/472e12f5b6505cc2fdd3b5e529b616b3a2d5fbf8

If not it throws:
`Rack::Lint::LintError: uppercase character in header name: Cache-Control`

cschlack:
Using rack < 3.0.0 requires all response header keys to be lowercase: https://github.com/rack/rack/commit/472e12f5b6505cc2fdd3b5e529b616b3a2d5fbf8

fixes #431

### DIFF
--- a/lib/sidekiq-scheduler/extensions/web.rb
+++ b/lib/sidekiq-scheduler/extensions/web.rb
@@ -10,5 +10,5 @@ if Sidekiq::VERSION >= '6.0.0'
   Sidekiq::Web.use Rack::Static, urls: ['/stylesheets-scheduler'],
                                  root: ASSETS_PATH,
                                  cascade: true,
-                                 header_rules: [[:all, { 'Cache-Control' => 'public, max-age=86400' }]]
+                                 header_rules: [[:all, { 'cache-control' => 'public, max-age=86400' }]]
 end


marcelolx:
Thank you!

lib/sidekiq-scheduler/extensions/web.rb
```diff
@@ -10,5 +10,5 @@
   Sidekiq::Web.use Rack::Static, urls: ['/stylesheets-scheduler'],
                                  root: ASSETS_PATH,
                                  cascade: true,
-                                 header_rules: [[:all, { 'Cache-Control' => 'public, max-age=86400' }]]
+                                 header_rules: [[:all, { 'cache-control' => 'public, max-age=86400' }]]
```
pbstriker38, line 13: Should this also have `, no-cache="Set-Cookie"`?

https://github.com/sidekiq/sidekiq/issues/5936

mperham: It shouldn’t need this at all. Extensions should be able to add asset paths and reuse the default Rack::Static instance that Sidekiq installs.

marcelolx: hmm interesting, I will check this when I have some time!

PR #434, opened 2023-04-14
Provide default for rufus scheduler

samsm:
Ran into this: https://github.com/sidekiq-scheduler/sidekiq-scheduler/issues/426

I toyed with the existing tests but didn't figure out a way to slip this in that I was happy about. So we have what we have here.

### DIFF
--- a/lib/sidekiq-scheduler/scheduler.rb
+++ b/lib/sidekiq-scheduler/scheduler.rb
@@ -58,7 +58,7 @@ module SidekiqScheduler
       self.dynamic = config.dynamic?
       self.dynamic_every = config.dynamic_every?
       self.listened_queues_only = config.listened_queues_only?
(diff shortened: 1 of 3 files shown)


samsm:
LOL I give up

marcelolx:
why? facing some problem with the test?

samsm:
Yeah was having a hard time writing a failing test. Stumped me

marcelolx:
But did you test that your change solves the problem it was trying to solve? We don't have a perfect coverage, so I wouldn't mind if we can't have a test for this line in particular

marcelolx:
Got the test working @samsm. It was just missing the `require 'sidekiq/capsule'`, but I think this probably is not going to solve the problem alone, this seems to be related to https://github.com/sidekiq-scheduler/sidekiq-scheduler/issues/419

samsm:
Thanks for tracking that down!

PR #446, opened 2023-06-15
Change cache-control to private

t27duck:
As mentioned in https://github.com/sidekiq/sidekiq/issues/5936 the public cache control could leak information when used with CloudFront or other CDNs. It was decided to change the cache setting in Sidekiq web's static assets to private.

This changes the gem's cache-control settings to match the recently released Sidekiq's.

### DIFF
--- a/lib/sidekiq-scheduler/extensions/web.rb
+++ b/lib/sidekiq-scheduler/extensions/web.rb
@@ -9,4 +9,4 @@ Sidekiq::Web.locales << File.expand_path("#{File.dirname(__FILE__)}/../../../web
 Sidekiq::Web.use Rack::Static, urls: ['/stylesheets-scheduler'],
                                root: ASSETS_PATH,
                                cascade: true,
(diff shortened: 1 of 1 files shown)


t27duck:
@marcelolx - I'll probably merge this myself sometime Friday evening or Saturday (US East time) unless someone thinks this is a horrible idea.

marcelolx:
I am good with it if that solves the problem, I had in mind to try to remove this entirely because of what Mike mentioned here https://github.com/sidekiq-scheduler/sidekiq-scheduler/pull/432#discussion_r1214971860, so we wouldn't have to worry about it anymore

But if that isn't as simple as doing this, we can move forward with it and revisit it

t27duck:
Let's go with this for now (to stop the... "bleeding"? for those who would be affected) and revist doing what Mike mentioned later then. 👍

marcelolx:
Sounds good to me, created an issue for it so I don't forget about it 🙈 https://github.com/sidekiq-scheduler/sidekiq-scheduler/issues/447

PR #455, opened 2023-10-21
Fix/cron jobs run more than once

muk-ai:
@marcelolx 
Thank you for maintaining a great gem.

I made this PR with reference to your PR. (https://github.com/sidekiq-scheduler/sidekiq-scheduler/pull/396) 
In your idea you were trying to calculate the scheduled time for cron, but I noticed that the jobs in rufus-scheduler already have that information. That is the [previous_time](https://github.com/jmettraux/rufus-scheduler#previous_time).
I rewrote it to use previous_time for cron jobs and added a test. I would be happy to have it merged if there is no problem.

### DIFF
--- a/lib/sidekiq-scheduler/scheduler.rb
+++ b/lib/sidekiq-scheduler/scheduler.rb
@@ -256,7 +256,12 @@ module SidekiqScheduler
       options = options.merge({ :job => true, :tags => [name] })
 
       rufus_scheduler.send(interval_type, schedule, options) do |job, time|
-        idempotent_job_enqueue(name, time, SidekiqScheduler::Utils.sanitize_job_config(config)) if job_enabled?(name)
+        if job_enabled?(name)
+          # rufus-scheduler returns the current time. For CronJob, use the job scheduled time.
+          time = job.previous_time if job.is_a?(Rufus::Scheduler::CronJob)
+
+          idempotent_job_enqueue(name, time, SidekiqScheduler::Utils.sanitize_job_config(config))
+        end
       end
     end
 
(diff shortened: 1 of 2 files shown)


marcelolx:
@muk-ai I am not sure this solves the issue, we already use the `previous_time` PR as well https://github.com/sidekiq-scheduler/sidekiq-scheduler/pull/396/files#diff-0ddd2bcc9bd29aac261acdd76530920c85887c269cc29e3f054e465b4ac8699eR355, but don't rely solely on it... 

I added this to my TODO list and will take a better look when I have more time and see if just this solves the problem or not. Thanks for opening the PR, though. Any help is always welcome!

muk-ai:
Oops, you were already using `previous_time` and trying to take some further care of it.
I will wait for your PR. You can refer to my test code if you like.

muk-ai:
@marcelolx 
Sorry, I did not understand how busy you were.

Let me explain about my motivation.
To completely prevent the sidekiq scheduler from running multiple cron jobs, the server must be configured as follows.
image

I would like to do that as follows.
image

There are three advantages to doing this.
- The scheduler becomes redundant.
- No longer have to run two differently configured sidekiqs.
- It prevents the users from unknowingly having multiple job runs.

So I would like you to check my PR again when you have the time, of course.
If you are unsure whether it will work well, you can make it an optional experimental feature. Then you can release it and we can try it out to see if it works in our production.

marcelolx:
Thanks @muk-ai! Would you have some time to look at #463? I see you just used `previous_time`, but that might not work if Rufus runs the jobs precisely at the time it is supposed to run — i.e. run every minute, if it runs exactly at 12:00:00 we would return 11:59:00 even though it could have already run previously at this time

Let me know what you think (feel free to continue the discussion in #463) and then we can work together to get this finally in the library 🙏

marcelolx:
Closing in favor of #463

muk-ai:
https://github.com/sidekiq-scheduler/sidekiq-scheduler/pull/463 This code is better. thanks!

PR #458, opened 2023-12-13
Update tilt dependency (or at least increase compatibility)

brunoocasali:
Hi team. Thanks for this fantastic work!

I will adopt this gem in an internal project here at meilisearch, but we use tilt internally in a newer version. 
Since I ran the test suite with the new version and it didn't break anything, I guess it wouldn't hurt to increase the compatibility range in the scheduler.

So this PR increases the compatibility range to accept any tilt version less than 3.


marcelolx:
Thanks @brunoocasali! The build seems to be broken for ruby 3+, but I will check that later, doesn't seem to be caused by your changes

brunoocasali:
Thank you 🤘

PR #463, opened 2024-03-07
Fix cron jobs timing

afrase:
Hey folks!

So I recently had some free time and decided to revisit this issue I had previously fixed at the company I was working for back in 2019.

I moved the calculation method into the `Utils` module to make testing easier and added lots of comments explaining what was going on. Added tests for the method since testing this issue through Rufus is incredibly difficult.

Sorry @marcelolx for never getting back to you on this.

This would close #396 #276 and #455

### DIFF
--- a/lib/sidekiq-scheduler/utils.rb
+++ b/lib/sidekiq-scheduler/utils.rb
@@ -141,5 +141,44 @@ module SidekiqScheduler
     def self.update_job_last_time(name, last_time)
       SidekiqScheduler::RedisManager.set_job_last_time(name, last_time) if last_time
     end
+
+    # Try to figure out when the cron job was supposed to run.
+    #
+    # Rufus calls the scheduler block with the current time and not the time the block was scheduled to run.
+    # This means under certain conditions you could have a job get scheduled multiple times because `time.to_i` is used
+    # to key the job in redis. If one server is under load and Rufus tries to run the jobs 1 seconds after the other
+    # server then the job will be queued twice.
+    # This method essentially makes a best guess at when this job was supposed to run and return that.
+    #
+    # @param [Fugit::Cron] cron
+    # @param [Time] time
+    #
+    # @return [Time]
+    def self.calc_cron_run_time(cron, time)
+      time = time.round # remove sub seconds to prevent rounding errors.
+      next_t = cron.next_time(time).utc
+      previous_t = cron.previous_time(time).utc
+      # The `time` var is some point between `previous_t` and `next_t`.
+      # Figure out how far off we are from each side in seconds.
+      next_diff = next_t - time
+      previous_diff = time - previous_t
+
+      if next_diff == previous_diff
+        # `next_diff` and `previous_diff` can be equal in two scenarios.
+        # 1. The `time` is exactly when the job was scheduled to run.
+        # 2. The `time` is exactly between `next_t` and `previous_t`, down to the exact second.
+        #
+        # The odds of 2 happening are basically zero but it is possible.
+        # So in scenario 2 it's possible to return the wrong time. For example if the job runs every minute and `time`
+        # is 15:25.30 then the returned time will be 15:25.30 instead of 15:25.0 (if rounding down).
+        time
+      elsif next_diff > previous_diff
+        # We are closer to the previous run time so return that.
+        previous_t
+      else
+        # We are closer to the next run time so return that.
+        next_t
+      end
+    end
   end
 end
--- a/lib/sidekiq-scheduler/utils.rb
+++ b/lib/sidekiq-scheduler/utils.rb
@@ -164,14 +164,9 @@ module SidekiqScheduler
       previous_diff = time - previous_t
 
       if next_diff == previous_diff
-        # `next_diff` and `previous_diff` can be equal in two scenarios.
-        # 1. The `time` is exactly when the job was scheduled to run.
-        # 2. The `time` is exactly between `next_t` and `previous_t`, down to the exact second.
-        #
-        # The odds of 2 happening are basically zero but it is possible.
-        # So in scenario 2 it's possible to return the wrong time. For example if the job runs every minute and `time`
-        # is 15:25.30 then the returned time will be 15:25.30 instead of 15:25.0 (if rounding down).
-        time
+        # In the event `time` is exactly between `previous_t` and `next_t` the diff will not be equal to
+        # `cron.rough_frequency`. In that case we round down.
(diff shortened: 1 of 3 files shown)


lib/sidekiq-scheduler/utils.rb
```diff
@@ -141,5 +141,44 @@ def self.update_job_next_time(name, next_time)
...
+        # So in scenario 2 it's possible to return the wrong time. For example if the job runs every minute and `time`
+        # is 15:25.30 then the returned time will be 15:25.30 instead of 15:25.0 (if rounding down).
+        time
+      elsif next_diff > previous_diff
+        # We are closer to the previous run time so return that.
+        previous_t
+      else
+        # We are closer to the next run time so return that.
+        next_t
+      end
```
afrase, line 181: So scenario 2 can actually be fixed but checking `cron.rough_frequency` against `next_diff`/`previous_diff`. If the diffs are equal AND `cron.rough_frequency` is equal to the diff then `time` is exactly when the job was scheduled to run. The problem with this is, what time do you return when exactly in the middle?

I think that decision is up to the maintainers and why I didn't go that route.

Here is how rounding down could be implemented.
```suggestion
      if next_diff == previous_diff
        # In the event `time` is exactly between `previous_t` and `next_t` the diff will not be equal to
        # `cron.rough_frequency`. In that case we round down.
        cron.rough_frequency == next_diff ? time : previous_t
      elsif next_diff > previous_diff
        # We are closer to the previous run time so return that.
        previous_t
      else
        # We are closer to the next run time so return that.
        next_t
      end
```

marcelolx:
Thanks for putting this together @afrase! I had a stab at this in https://github.com/sidekiq-scheduler/sidekiq-scheduler/pull/396 but never did get back to it

I will set some time aside to look into the comments you left in the code 🙏 Looking forward to incorporate this

lib/sidekiq-scheduler/utils.rb
marcelolx: I think returning the previous time in that scenario makes sense because this might only happen if the server is on high load, so in theory, the job never will be running before its scheduled time, only after, so if it falls exactly in the middle, it is very likely it was meant to run in the previous time...

That makes me wonder if there is any scenario in which we want to return `next_t`? Have you seen it happen that the scheduler enqueued the job a few seconds or milliseconds before the supposed time it should run?

I am good to move forward with this path here, just want to make sure I understand the scenario for the `next_t`

And sorry for taking so long to take a look on this PR

marcelolx: I think returning the previous time in that scenario makes sense because this might only happen if the server is on high load, so in theory, the job never will be running before its scheduled time, only after, so if it falls exactly in the middle, it is very likely it was meant to run in the previous time...

That makes me wonder if there is any scenario in which we want to return `next_t`? Have you seen it happen that the scheduler enqueued the job a few seconds or milliseconds before the supposed time it should run?

I am good to move forward with this path here, just want to make sure I understand the scenario for the `next_t`

And sorry for taking so long to take a look on this PR

afrase: > I think returning the previous time in that scenario makes sense because this might only happen if the server is on high load, so in theory, the job never will be running before its scheduled time, only after, so if it falls exactly in the middle, it is very likely it was meant to run in the previous time...

Yes, I agree returning the previous time when exactly in the middle makes the most sense.


> That makes me wonder if there is any scenario in which we want to return `next_t`? Have you seen it happen that the scheduler enqueued the job a few seconds or milliseconds before the supposed time it should run?

The `next_t` is more of a hypothetical scenario just to have all bases covered. I have never seen a job run before it's scheduled and I'm not sure if it's even possible with how the scheduler works.

marcelolx: > Yes, I agree returning the previous time when exactly in the middle makes the most sense.

Mind adding this change then and changing the test to cover it? I think we can merge it then!

> The next_t is more of a hypothetical scenario just to have all bases covered. I have never seen a job run before it's scheduled and I'm not sure if it's even possible with how the scheduler works.

I see, yeah, I think we can move forward as it is! Only scenario I see that it could become an issue if it is for a job that runs every minute and the load is so high that rufus only triggers it after 30s+ it was supposed to run, then we will enqueue it for the next minute, but it is hard to imagine someone will have a load so high that this would happen. 

Thanks @afrase!

afrase: Thats been updated.

Yeah something would have to be very wrong to get a job trying to schedule 30+ seconds after scheduled.
The company where I fixed this heavily used scheduled jobs and had 2B+ jobs ran and I think the most delayed I saw was 3 seconds.

marcelolx: Perfect, thanks @afrase! this looks good!

marcelolx:
I don't have a specific one, but essentially, since we don't use any global lock mechanism to push a job to Redis, there is always the possibility that two different hosts could push the same job at the same time. That though, became very unlikely with this change.

By global lock, I mean a lock that each host has to acquire to then push the job to the queue, something in similar to this https://github.com/sidekiq-scheduler/sidekiq-scheduler/issues/332#issuecomment-1213061181

michelson:
in an scenario where I have multiple sidekiq hosts but only have cron running on a single machine, should I worry about the message being executed multiple times by other hosts?

muk-ai:
@afrase @marcelolx 
You've done a great job. Thanks!
When the new version of the gem is released, we will use it in our products.

muk-ai:
I was assuming intervals like once a day, but this PR seems to take into account shorter intervals like 1 minute intervals. Very nice.

marcelolx:
@muk-ai fyi v5.0.4 has been released with those changes here

muk-ai:
@marcelolx 
Thanks for the 5.0.4 release.
We updated it immediately. It is working fine, but I have not yet been able to generate an event that delays job execution by more than 1 second when the server load is high.

```log
@timestamp
2024-07-17 04:58:00.109
2024-07-17 04:57:00.049
2024-07-17 04:56:00.105
: :
2024-07-16 05:03:00.111
2024-07-16 05:02:00.170
2024-07-16 05:01:00.099
```

I am sure that the code prevents multiple executions. Thanks again!

y-yagi:
Hi. I'm doubting this is causing the unexpected behavior. I've submitted the issue #482. Could you check it?

marcelolx:
@muk-ai thanks for letting me know!

@y-yagi thanks for the report, already merged!

PR #464, opened 2024-03-07
Use Redis MULTI command

afrase:
This is another change that I think helped eliminate the multi job issue in addition to #463.

From the Redis [docs](https://redis.io/docs/interact/transactions/)
> All the commands in a transaction are serialized and executed sequentially. A request sent by another client will never be served in the middle of the execution of a Redis Transaction. This guarantees that the commands are executed as a single isolated operation.

The `multi` block form also pipelines the commands so this is just making the command a transaction.


marcelolx:
thank you!

Issue #284, opened 2019-08-09
Do dynamic 'every' schedules die when sidekiq dies?


jcavalieri:
I'm seeing after a deploy or restart of sidekiq that our dynamic 'every' schedules are gone and no longer working.

Is this correct behavior and is there a way to recover?

Thanks,
-John

jcavalieri:
FYI,
I'm see that this persists in redis:
```
r.hgetall 'sidekiq-scheduler:next_times'
=> {"schedule-231::fixture-1::frequency-1h::2019-08-09T03:04:00.000-04:00_2019-08-09T10:03:59.000-04:00"=>"2019-08-09 03:04:00 -0400",
 "schedule-233::fixture-1::frequency-10m::2019-08-09T10:04:00.000-04:00_2019-08-10T03:03:59.000-04:00"=>"2019-08-09 10:44:00 -0400",
 "schedule-233::fixture-1::frequency-10m::2019-08-08T10:04:00.000-04:00_2019-08-09T03:03:59.000-04:00"=>"2019-08-09 00:24:00 -0400",
 ""=>"2019-08-09 10:40:51 -0400"}
```

jcavalieri:
And don't know if this is helpful, but these are all the redis keys for sidekiq-scheduler that I see.
```[29] pry(main)> puts r.keys.select{|k|k.include?('sidekiq-scheduler:')}
sidekiq-scheduler:last_times
sidekiq-scheduler:pushed:schedule-233::fixture-1::frequency-10m::2019-08-09T10:04:00.000-04:00_2019-08-10T03:03:59.000-04:00
sidekiq-scheduler:pushed:schedule-233::fixture-1::frequency-10m::2019-08-08T10:04:00.000-04:00_2019-08-09T03:03:59.000-04:00
sidekiq-scheduler:next_times
=> nil```

rusucosmin:
+1

Issue #227, opened 2017-11-07
Dynamic schedule created, then deleted during the boot up


zedtux:
I'm trying to use the [dynamic-schedule](https://github.com/moove-it/sidekiq-scheduler#dynamic-schedule) provided by this gem in order to create a recurring job where the cron is coming from the config of my application.

In the Sidekiq initialiser I have something like this:

```ruby
# ... sidekiq intialize code here ...

if Sidekiq.server?
  subscription_payments_job = 'subscription_payments'
  if Sidekiq.get_schedule(subscription_payments_job)
    Rails.logger.info '[Sidekiq] Scheduled job ' \
                      "#{subscription_payments_job.inspect} found, " \
                      'removing it ...'
    Sidekiq.remove_schedule(subscription_payments_job)
  end

  Rails.logger.info '[Sidekiq] Re-creating the ' \
                    "#{subscription_payments_job.inspect} job with " \
                    "billing_day #{Rails.configuration.billing_day} ..."

  Sidekiq.set_schedule(subscription_payments_job, {
                         class: 'CollectEntitySubscriptionPaymentJob',
                         description: 'Create transaction and get the money',
                         queue: 'default',
                         cron: "0 9 #{Rails.configuration.billing_day} * *"
                       })

  if Sidekiq.get_schedule(subscription_payments_job)
    Rails.logger.info '[Sidekiq] Recurring job ' \
                      "#{subscription_payments_job.inspect} successfully " \
                      'created!'
  else
    Rails.logger.warn '[Sidekiq] Recurring job ' \
                      "#{subscription_payments_job.inspect} couldn't be " \
                      'created :('
  end
end
```

This creates the recurring job and then delete it (I guess in the boot up phase of Sidekiq, the sidekiq-scheduler calls the cleaning method before the load one).

So apparently I'm too early in the boot phase. What shall I do in order to get my job not deleted?

zedtux:
Well ... I just figured it out apparently (maybe you can confirm? 😄):

```ruby
Sidekiq.configure_server do |config|
  # ... sidekiq intialize code here ...

  config.on(:startup) do
    # My code here
  end
end
```

And now it's working.

Issue #188, opened 2017-05-16
Remember dynamic schedule on deploys


sineed:
Hi! I just started using your gem. I found it very useful because it allows to make dynamic schedules.

I also found that it does not remember the last schedule if I restart my Sidekiq process. It usually happens on deployment.

I thought about hooking into deployment process to create a file and write to it the latest `Sidekiq.get_schedule` value. After successful start of Sidekiq I need to do the same thing as described [here](https://github.com/moove-it/sidekiq-scheduler#load-the-schedule-from-a-different-file).

So in the end I will have the following configuration for Sidekiq:

```ruby
require "sidekiq"
require "sidekiq/scheduler"

RELATIVE_PATH = "path_to_schedule_file.yml"

def get_schedule
  schedule   = Sidekiq.get_schedule
  next_times = Redis.current.hgetall(Sidekiq::Scheduler.next_times_key)

  next_times.each do |job_key, first_at|
    next unless schedule[job_key]
    schedule[job_key]["every"] << { "first_at" => first_at }
  end

  schedule
end

Sidekiq.configure_server do |config|
  config.on(:startup) do
    file_path = File.expand_path(RELATIVE_PATH, __FILE__)
    if File.exist?(file_path)
      File.open(file_path) do |f|
        Sidekiq.schedule = YAML.load_file(f)
        Sidekiq::Scheduler.reload_schedule!
      end
    end
  end

  config.on(:quiet) do
    file_path = File.expand_path(RELATIVE_PATH, __FILE__)
    file = File.new(file_path, "w+")
    file << get_schedule.to_yaml
    file.close
  end

  config.on(:shutdown) do
    file_path = File.expand_path(RELATIVE_PATH, __FILE__)
    file = File.new(file_path, "w+")
    file << get_schedule.to_yaml
    file.close
  end
end

```

This code is wrong, it just shows the basic idea.

Maybe my findings can be useful for someone else.

--
UPD. I've updated configuration example, now it is correct and is used in my project. I want to point that for me it is also important to remember the next time a job will be enqueued

rrrcompagnoni:
I'm getting this issue too. It seems that Rufus Scheduler is being reset every time that I restart my Sidekiq worker in dynamic mode. I'd like to know if I can override this default behavior to keep my scheduled jobs without any side-effects after that.

ladiadeniran:
Nice one keeping this out here. Our workaround involved writing a job that runs on starting sidekiq.

[...]

Issue #125, opened 2016-08-26
Pure dynamic schedules not working, Schedule empty! Set Sidekiq.schedule


avongluck-r1soft:
I'm playing with the idea of purely dynamic schedules... however the dynamic schedules never fire.

```
Scheduling environment polling job for -fa1b-4853... (set_schedule)
Scheduling environment polling job for -89c5-40a4... (set_schedule)
Scheduling environment polling job for -8367-4e06... (set_schedule)
(Sidekiq::Scheduler.reload_schedule!)
2016-08-26T18:38:07.849Z 10005 TID-7ltrg INFO: Reloading Schedule
2016-08-26T18:38:07.852Z 10005 TID-7ltrg INFO: Loading Schedule
2016-08-26T18:38:07.852Z 10005 TID-7ltrg INFO: Schedule empty! Set Sidekiq.schedule
2016-08-26T18:38:07.856Z 10005 TID-7ltrg INFO: Schedules Loaded
```

get_all_schedules shows the dynamic scheduled jobs... however they never fire and I see the error above.

init:
```
require 'sidekiq/scheduler'

Sidekiq.configure_server do |config|
  config.on(:startup) do
    Sidekiq::Scheduler.dynamic = true
  end
end
```

artem-russkikh:
If you plan queue jobs from sidekiq thread, try to set dynamic in initializer like so

```ruby
# config/initializers/sidekiq_scheduler.rb

require 'sidekiq/scheduler'
Sidekiq::Scheduler.dynamic = true
```

I queue jobs only in Unicorn thread:

```ruby
# config/initializers/sidekiq_scheduler.rb

require 'sidekiq/scheduler'

Sidekiq::Scheduler.dynamic = true

if Rails.env.production? && (defined?(Rails::Server) || defined?(Unicorn))
  Sidekiq::Scheduler.enabled = true
  scheduler_options = {
    dynamic: true,
    enabled: true,
    schedule: nil,
    listened_queues_only: nil
  }
  schedule_manager = SidekiqScheduler::Manager.new(scheduler_options)
  schedule_manager.start
elsif Rails.env.test?
  Sidekiq::Scheduler.enabled = true
else
  Sidekiq::Scheduler.enabled = false
end
```

[...]

PR #467, opened 2024-05-17
Don't clear "schedule" key on boot in dynamic schedule mode

Floppy:
There have been a number of issues around dynamic jobs on boot (e.g. #188, #227, #284, #125), which I think all have a common root cause.

During Manager#initialize, the schedule is loaded, even if that schedule is empty, and even in dynamic mode. See https://github.com/sidekiq-scheduler/sidekiq-scheduler/blob/9da5f130072ae8bf8f500862f44789029234b84e/lib/sidekiq-scheduler/manager.rb#L16

This causes the `schedule` key to be removed from Redis as it's now empty, so when we get to https://github.com/sidekiq-scheduler/sidekiq-scheduler/blob/9da5f130072ae8bf8f500862f44789029234b84e/lib/sidekiq-scheduler/scheduler.rb#L85, the schedule cannot be loaded as it's already gone.

This PR simply disables that initial setting of the schedule in dynamic mode.

Obviously this might affect existing running systems... not sure what to do about that.

### DIFF
--- a/lib/sidekiq-scheduler/manager.rb
+++ b/lib/sidekiq-scheduler/manager.rb
@@ -13,7 +13,7 @@ module SidekiqScheduler
 
       @scheduler_instance = SidekiqScheduler::Scheduler.new(config)
       SidekiqScheduler::Scheduler.instance = @scheduler_instance
-      Sidekiq.schedule = config.schedule if @scheduler_instance.enabled
+      Sidekiq.schedule = config.schedule if @scheduler_instance.enabled && !@scheduler_instance.dynamic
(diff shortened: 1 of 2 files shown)


marcelolx:
Thanks @Floppy — once I have more time I will look into this but I agree it might affect existing systems, so we might need to do a major release

marcelolx:
This sounds correct @Floppy — so it will make it to the next beta of v6.

I still want to test some things to ensure it behaves as we want.

Also, sorry for taking so long to come back to this 😓

Issue #468, opened 2024-06-18
Implement `to_hash` in SidekiqScheduler::Manager to avoid Stack level too deep error.


yiyenene:
Hi.

I faced Stack level too deep error when sidekiq dumps log of its config.
It was caused by my logger using ActiveSupport's `as_json`.

This gem sets `SidekiqScheduler::Manager` instance to sidekiq's config and Manager instance has sidekiq's config inside through SidekiqManager::Config instance. 
There is a circular reference. (and also is in the Rufus::Scheduler instance.)

ActiveSupoort's `as_json` traverses `instance_values` unless it responds to `to_hash`. 
So, how about implement `to_hash` in SidekiqScheduler::Manager to avoid unintended exposing instance values and stack level too deep error ?

Would it be alright if I create the PR ?

ref 
- https://github.com/rails/rails/issues/51626
- https://github.com/sidekiq/sidekiq/issues/5920

I apologize if my English seems rude.

marcelolx:
@yiyenene, thanks for the report; I think this might also be related to this issue: https://github.com/sidekiq-scheduler/sidekiq-scheduler/issues/462 — A PR is definitely welcome, and your English isn't rude!

yiyenene:
Thank you for merging and fixing typo 🙇

Issue #462, opened 2024-03-04
SidekiqScheduler::Manager enormous inspect output


eapache-opslevel:
Trying to inspect an instance of `SidekiqScheduler::Manager` produces around 5k lines of output on our system because it holds references to a bunch of other complex objects and eventually starts printing large chunks of internal sidekiq and rails objects.

This is doubly annoying because with debug-logging enabled on sidekiq, the `SidekiqScheduler::Manager` instance is inspected/logged at sidekiq startup.

It's easy enough to monkey-patch with something like:

```ruby
module SidekiqScheduler
  class Manager
    def inspect
      "SidekiqScheduler::Manager"
    end
  end
end
```

But there should probably be *some* debugging information included, and I don't know what values are worth logging here.

eapache-opslevel:
I am not able to share this output, as the resulting dump of ruby objects includes a great deal of sensitive information about our code base, including e.g. our redis password. It is 2MB of text, so it is too large for me to manually redact sensitive information.

You should be able to reproduce for yourself by running `bundle exec sidekiq -v` on a sidekiq setup with sidekiq-scheduler. Any sidekiq version after 7.0.7 (https://github.com/sidekiq/sidekiq/issues/5822 is required) should print the config object when started with `-v` and will thus trigger the inspect of the scheduler-manager.

afrase:
I've had this same issue also because Sidekiq would print `Sidekiq.options` and the scheduler instance is in those options.
My notes say the problem is with the rufus scheduler having an array of rufus jobs which all have a reference to the rufus scheduler.

We fixed this by doing this in the initializer.

```ruby
module SidekiqScheduler
  class Scheduler
    def inspect
      "#<#{self.class}:0x#{object_id.to_s(16)}>"
    end
  end
end
```

PR #470, opened 2024-06-19
Implement `to_hash` method in Manager, Scheduler and Config classes.

yiyenene:
ref #468 

## What I did

- add `to_hash` method to `SidekiqScheduler::Manager`, `SidekiqScheduler::Scheduler` and `SidekiqScheduler::Config` classes.
- create `spec/sidekiq-scheduler/config_spec.rb` to test.
- add test in `spec/sidekiq-scheduler/scheduler_spec.rb` and `spec_sidekiq-scheduler/manager_spec.rb`.

Here is an example output of `SidekiqScheduler::Manager#to_hash`.
```ruby
{
  "scheduler" => {
    "scheduler_config" => {
      "enabled" => true, 
      "dynamic" => true, 
      "dynamic_every" => "5s", 
      "shedule" => {
        "current" => {
          "cron" => "* * * * *", 
          "class" => "SomeWorker", 
          "args" => "some_arg", 
          "queue" => "default"
        }
      }, 
      "listened_queues_only" => true, 
      "rufus_scheduler_options" => {}
    }
  }
}
```

All spec are passed in my local.

Unfortunately, this doesn't affect `inspect` behavior. We need more improvement to solve #462

### DIFF
--- a/lib/sidekiq-scheduler/config.rb
+++ b/lib/sidekiq-scheduler/config.rb
@@ -59,6 +59,17 @@ module SidekiqScheduler
       SidekiqScheduler::SidekiqAdapter.sidekiq_queues(sidekiq_config)
     end
 
+    def to_hash
+      {
+        enabled: enabled?,
+        dynamic: dynamic?,
+        dynamic_every: dynamic_every?,
+        shedule: schedule,
+        listened_queues_only: listened_queues_only?,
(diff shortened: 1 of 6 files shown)


marcelolx:
This is a good starting point, thanks @yiyenene!

PR #472, opened 2024-07-02
Use new web ui registration and stylesheet helpers

t27duck:
Sidekiq 7.3.0 and later added changes to how the web UI for plugins work.

- The registration has a first-class API to register a plugin.
  - See: https://github.com/sidekiq/sidekiq/blob/main/examples/webui-ext/lib/sidekiq-redis_info/web.rb#L20
  - The old registration path is still allowed, but will be removed in Sidekiq 8.0.
- New stylesheet and script helpers were added to allow linking to scripts and stylesheets for the plugin. They add nonce attributes to protect against XSS attacks.
  - See: https://github.com/sidekiq/sidekiq/blob/main/examples/webui-ext/readme.md
  - Sidekiq 8.0 and later will require nouce attributes for all of Sidekiq web.

To make this gem forwards compatible with 8.0...

- Added a new constant to check for Sidekiq 7.3 and later.
- For current and future Sidekiq releases, we'll use the new `style_tag` helper to output the `<link>` tag. Older Sidekiq versions will continue to use the hand-crafted tag as it stands today.
- For current and future Sidekiq releases, we'll use the newer registration API for the web UI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
       fail-fast: false
       matrix:
         redis-version: ["~> 4.2", "~> 5"]
-        ruby-version: ['2.7', '3.0', '3.1', '3.2', '3.3']
+        ruby-version: ['3.1', '3.2', '3.3']
         sidekiq-version: ['~> 6', '~> 7']
     services:
       redis:
--- a/web/locales/cs.yml
+++ b/web/locales/cs.yml
@@ -1,4 +1,5 @@
 cs:
+  "Recurring Jobs": Pravidelně opakované
   recurring_jobs: Pravidelně opakované
   name: Jméno
   description: Popis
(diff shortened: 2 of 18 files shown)


.github/workflows/ci.yml
```diff
@@ -11,7 +11,7 @@ jobs:
       fail-fast: false
       matrix:
         redis-version: ["~> 4.2", "~> 5"]
-        ruby-version: ['2.7', '3.0', '3.1', '3.2', '3.3']
+        ruby-version: ['3.1', '3.2', '3.3']
```
t27duck, line 14: Figured may as well since 3.1+ is the only Supported Rubies now.

web/locales/cs.yml
```diff
@@ -1,4 +1,5 @@
 cs:
+  "Recurring Jobs": Pravidelně opakované
```
t27duck, line 2: This was the only way I could get the tab title to work when using the locale switcher at the bottom of Sidekiq web UI. 🤷‍♂️

t27duck:
@marcelolx The gem as-is should be fine for Sidekiq 7.3 which was dropped recently, but I figure may as well jump on this.

marcelolx:
This is great @t27duck! I really appreciate this 🙏

PR #479, opened 2024-07-09
[POTENTIAL BREAKING CHANGE] Fix typo in `config#to_hash`

tagliala:
Hello, is this done on purpose?

Since this is a potential breaking change, I can preserve `shedule`, and that should be deprecated and removed in 6.0

---

Potential breaking change

Ref: #470


marcelolx:
Definitely not intentional, thanks for catching that

PR #480, opened 2024-07-09
Opt-in for MFA requirement

tagliala:
Make the gem more secure by requiring that all privileged operations by any of the owners require OTP.

Ref: https://guides.rubygems.org/mfa-requirement-opt-in/

---

Hello,

even if the numbers of downloads can implicitly require MFA to publish this gem, adding this metadata will add an explicit panel on the right sidebar of the gem.

This is also in Sidekiq and a lot of popular gems (which already require MFA)

Ref: sidekiq/sidekiq#5850

image


marcelolx:
Thanks @tagliala

Issue #482, opened 2024-07-18
Jobs don't correctly register when using the cron hour range since v5.0.4


y-yagi:
Since #463,  `sidekiq-scheduler` uses the diff between a previous job execution time and a next job execution time to decide a job time. But, I'm assuming this check doesn't work correctly when jobs don't run periodically.

For example, considering the case of the cron setting is `0 8-18 * * *`.

First, when the job execution time is `2024/07/18 08:00:00 JST`, `next_diff` and `previous_diff` aren't the same because the `previous_time` is 8 hours ago. So the next execution time(`2024/07/18 09:00:00 JST`) will be used as the key of the job.
https://github.com/sidekiq-scheduler/sidekiq-scheduler/blob/42c12c04045453dcda78c870733c344aedaa5c67/lib/sidekiq-scheduler/utils.rb#L166-L176

Next job will execute on `2024/07/18 09:00:00 JST`. In this case, `next_diff` and `previous_diff` are the same. So the  `2024/07/18 09:00:00 JST` will used as the key of the job. But this key has already used when the execution time is `2024/07/18 08:00:00 JST`. As a result, this job will not run. I think this isn't expected behavior.

y-yagi:
I created the PR to fix the issue #484.

PR #484, opened 2024-07-21
Only call SidekiqScheduler::Utils.calc_cron_run_time when it seems really necessary.

y-yagi:
As I described #482,  `SidekiqScheduler::Utils.calc_cron_run_time` only works with a job running periodically.
Also, `SidekiqScheduler::Utils.calc_cron_run_time` is only required jobs are execute frequently I think. So this PR changes to call it when it really needed.

Also, I added the test for `#load_schedule_job` when the cron is not periodically to prevent regression.

### DIFF
--- a/lib/sidekiq-scheduler/scheduler.rb
+++ b/lib/sidekiq-scheduler/scheduler.rb
@@ -267,7 +267,9 @@ module SidekiqScheduler
         if job_enabled?(name)
           conf = SidekiqScheduler::Utils.sanitize_job_config(config)
 
-          if job.is_a?(Rufus::Scheduler::CronJob)
+          # `SidekiqScheduler::Utils.calc_cron_run_time` only works with a job running periodically.
+          # Also, `SidekiqScheduler::Utils.calc_cron_run_time` is only required jobs are execute frequently. So only call it if really needed.
+          if job.is_a?(Rufus::Scheduler::CronJob) && job.cron_line.rough_frequency <= 60
             idempotent_job_enqueue(name, SidekiqScheduler::Utils.calc_cron_run_time(job.cron_line, time.utc), conf)
           else
             idempotent_job_enqueue(name, time, conf)
--- a/spec/sidekiq-scheduler/scheduler_spec.rb
+++ b/spec/sidekiq-scheduler/scheduler_spec.rb
@@ -582,8 +582,14 @@ describe SidekiqScheduler::Scheduler do
     let(:next_time_execution) do
       SidekiqScheduler::Store.hexists(SidekiqScheduler::RedisManager.next_times_key, job_name)
     end
+    let(:travel_to) { nil }
 
-    before { subject }
+    before do
+      Timecop.travel(travel_to) if travel_to
+      subject
+    end
+
+    after { Timecop.return if travel_to }
 
     context 'without a timing option' do
       it 'does not put the job inside the scheduled hash' do
@@ -654,6 +660,27 @@ describe SidekiqScheduler::Scheduler do
 
           it 'does not store the next time execution correctly' do
             expect(next_time_execution).not_to be
+          end
+        end
+
+        context 'when the cron is not periodically' do
+          before { Sidekiq.schedule = { job_name => config } }
+          after { SidekiqScheduler::Store.del(pushed_job_key) }
+
+          let(:config) { ScheduleFaker.cron_schedule('cron' => '0 8-18 * * *') }
+          let(:pushed_job_key) { SidekiqScheduler::RedisManager.pushed_job_key(job_name) }
+          let(:travel_to) { Timecop.travel(Time.local(2024, 7, 1, 7, 0, 0)) }
+
+          it 'register consecutive jobs correctly' do
+            expect {
+              Timecop.travel(60 * 60)
+              sleep 0.5
+            }.to change { SidekiqScheduler::Store.zrange(pushed_job_key, 0, -1).size }.by(1)
+
+            expect {
+              Timecop.travel(60 * 60)
+              sleep 0.5
+            }.to change { SidekiqScheduler::Store.zrange(pushed_job_key, 0, -1).size }.by(1)
           end
         end
       end


lib/sidekiq-scheduler/scheduler.rb
```diff
@@ -267,7 +267,9 @@ def new_job(name, interval_type, config, schedule, options)
         if job_enabled?(name)
           conf = SidekiqScheduler::Utils.sanitize_job_config(config)
 
-          if job.is_a?(Rufus::Scheduler::CronJob)
+          # `SidekiqScheduler::Utils.calc_cron_run_time` only works with a job running periodically.
+          # Also, `SidekiqScheduler::Utils.calc_cron_run_time` is only required jobs are execute frequently. So only call it if really needed.
+          if job.is_a?(Rufus::Scheduler::CronJob) && job.cron_line.rough_frequency <= 60
```
y-yagi, line 272: I do not have a strong basis for this figure, but I thought 60 was sufficient, since I think this problem only occurs in cases where a batch is run every minute.

y-yagi: One more note: `SidekiqScheduler::Utils.calc_cron_run_time` returns a time that zone is converted to `utc`. To keep the time zone, I think it's good to avoid calling it if possible.

marcelolx:
Thanks @y-yagi! I am good with 60 too

afrase:
@y-yagi @marcelolx Does this mean only jobs running every 60 seconds will have the time corrected? The original fix for this came about because of daily / weekly jobs that would run multiple times under load.

I'm wondering if maybe removing the `rough_frequency` rounding stuff and going back to my original algorithm would fix these cases. 🤔

marcelolx:
@afrase I didn't realize it was for jobs running daily / weekly (so yeah, this would only correct time for jobs running every 60 seconds) 

Yeah, maybe we just revert that piece of the rough time and keep your first original version without it (that would fix this issue here without limiting it to every 60 seconds I think)

y-yagi:
@afrase 

> I'm wondering if maybe removing the rough_frequency rounding stuff and going back to my original algorithm would fix these cases

Do you mean to code the following?

```diff
diff --git a/lib/sidekiq-scheduler/utils.rb b/lib/sidekiq-scheduler/utils.rb
index cccb01c..aca12e5 100644
--- a/lib/sidekiq-scheduler/utils.rb
+++ b/lib/sidekiq-scheduler/utils.rb
@@ -164,15 +164,13 @@ module SidekiqScheduler
       previous_diff = time - previous_t

       if next_diff == previous_diff
-        # In the event `time` is exactly between `previous_t` and `next_t` the diff will not be equal to
-        # `cron.rough_frequency`. In that case we round down.
-        cron.rough_frequency == next_diff ? time : previous_t
+        time
       elsif next_diff > previous_diff
         # We are closer to the previous run time so return that.
-        previous_t
+        time - previous_diff
       else
         # We are closer to the next run time so return that.
-        next_t
+        time + next_diff
       end
     end
   end
```

If so, no. This code still has the issue that I encountered.

muk-ai:
I was looking into this matter and noticed a difference in the behavior of the `previous_time` returned by rufus-scheduler.

```ruby
scheduler = Rufus::Scheduler.new
scheduler.schedule_cron '1-3 * * * * *' do |job|
  puts "now: #{Time.now.inspect}"
  puts "previous_time1: #{job.previous_time}"
  puts "previous_time2: #{job.cron_line.previous_time}"
  puts "---"
end
```

output
```
now: 2024-07-23 13:15:01.294697632 +0900
previous_time1: 2024-07-23 13:15:01 +0900 <= ★
previous_time2: 2024-07-23 13:14:03 +0900 <= ★
---
now: 2024-07-23 13:15:02.196055795 +0900
previous_time1: 2024-07-23 13:15:02 +0900
previous_time2: 2024-07-23 13:15:01 +0900
---
now: 2024-07-23 13:15:03.097378346 +0900
previous_time1: 2024-07-23 13:15:03 +0900
previous_time2: 2024-07-23 13:15:02 +0900
---
```

`job.previous_time` returns the last execution time, whereas `job.cron_line.previous_time` returns the one previous execution time.

So, how about modifying it as follows?
```ruby
if job.is_a?(Rufus::Scheduler::CronJob)
  if job.cron_line.rough_frequency <= 60
    idempotent_job_enqueue(name, SidekiqScheduler::Utils.calc_cron_run_time(job.cron_line, time.utc), conf)
  else
    # No exact calculation is performed. Just discard seconds.
    idempotent_job_enqueue(name, SidekiqScheduler::Utils.discard_seconds(job.previous_time), conf)
  end
else
  idempotent_job_enqueue(name, time, conf)
end
```

y-yagi:
@muk-ai 
In your example, `rough_frequency` is under 60. So `SidekiqScheduler::Utils.discard_seconds` will not call, just call `SidekiqScheduler::Utils.calc_cron_run_time` as same as the current release's gem I assume.

```ruby
cron = Fugit.parse_cron('1-3 * * * * *')
cron.rough_frequency #=> 1
```

muk-ai:
> @muk-ai In your example, `rough_frequency` is under 60. So `SidekiqScheduler::Utils.discard_seconds` will not call, just call `SidekiqScheduler::Utils.calc_cron_run_time` as same as the current release's gem I assume.
> 
> ```ruby
> cron = Fugit.parse_cron('1-3 * * * * *')
> cron.rough_frequency #=> 1
> ```

That example is just for checking behaviour.

The use case I am thinking of for the job would be a daily or weekly job.
When the server is under high load, the problem is that the timing of job execution is delayed by a few seconds. Then the “time” used for the key does not match and the job is executed on multiple hosts.

afrase:
@y-yagi @muk-ai So I just discovered the `Fugit::Cron#match?` method.

From my quick testing it looks like if we just add a guard clause it fixes this.
```ruby
def self.calc_cron_run_time(cron, time)
  time = time.round # remove sub seconds to prevent rounding errors.
  return time if cron.match?(time) # Exact match, return the time.

  next_t = cron.next_time(time).utc
  previous_t = cron.previous_time(time).utc
  # The `time` var is some point between `previous_t` and `next_t`.
  # Figure out how far off we are from each side in seconds.
  next_diff = next_t - time
  previous_diff = time - previous_t

  if next_diff == previous_diff
    # In the event `time` is exactly between `previous_t` and `next_t` the diff will not be equal to
    # `cron.rough_frequency`. In that case we round down.
    cron.rough_frequency == next_diff ? time : previous_t
  elsif next_diff > previous_diff
    # We are closer to the previous run time so return that.
    previous_t
  else
    # We are closer to the next run time so return that.
    next_t
  end
end
```

muk-ai:
@afrase
It's elegant.
It may be better to use `floor` rather than `round`.

afrase:
> @afrase It's elegant. It may be better to use `floor` rather than `round`.

Yeah `floor` would probably be better since it truncates.

y-yagi:
@afrase 

I can't guarantee it, but using a future time sounds a little dangerous to me. 
Isn't it enough to just use `previous_time` when time isn't matched?

```ruby
def self.calc_cron_run_time(cron, time) 
  cron.match?(time) ? time : cron.previous_time(time)
end 
```

afrase:
@y-yagi 

The reason `next_time` is used is to try to cover all hypothetical scenarios. 
It may actually be impossible for Rufus to call a job _before_ it was scheduled to run but I wanted to cover that case. The `next_time` is only ever used as a last resort for that reason.

I'm working on a PR to incorporate the aperiodic time's fix. I'm also looking more closely at the use of time zones and how those could impact the result after you pointed out the use of `#utc`. The reason the for `#utc` is to convert `EtOrbi::EoTime` to `Time` for use with `Time#-`. I'm kind of leaning to just always using UTC to simplify things but I'm not sure yet.

PR #486, opened 2024-07-22
Test against all supported Ruby versions

### LINKED ISSUE #477: Keep testing against 2.7 and 3.0 in version 5.0.x to avoid breaking changes at patch level
Hello,

this change

https://github.com/sidekiq-scheduler/sidekiq-scheduler/pull/472/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR14

introduced in #472 may break backward compatibility in a patch release if someone starts using new Ruby 3.1 features, like `Array#intersect?` https://rubyapi.org/3.1/o/array#method-i-intersect-3F

This happened in the past, especially between 2.2 and 2.3 and the safe navigation operation or the squiggly heredoc, so some gems introduced breaking changes at patch level without noticing, having the developers to lock their gemfile to 'x.y.z', which is ok since it is a "legacy app" but can be avoided

This is one of the latest examples I remember: thoughtbot/factory_bot#1614, but there are many others

tagliala:
Prevents introduction of breaking changes at patch or minor version level

- [x] https://github.com/sidekiq/sidekiq/issues/6371
- [x] Pending Sidekiq release 

Close #477

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
       fail-fast: false
       matrix:
         redis-version: ["~> 4.2", "~> 5"]
-        ruby-version: ['3.1', '3.2', '3.3']
+        ruby-version: ['2.7', '3.0', '3.1', '3.2', '3.3']
         sidekiq-version: ['~> 6', '~> 7']
     services:
       redis:


tagliala:
Hello, 

I forgot to complete this PR. Sidekiq 7.3.1 was released a long time ago and now all is green

PR #487, opened 2024-07-30
Fix scheduling for aperiodic cron jobs

afrase:
This PR should fix the issues discussed in https://github.com/sidekiq-scheduler/sidekiq-scheduler/pull/484

Thanks @y-yagi @muk-ai and @marcelolx

If anyone has opinions on the use of local time vs UTC let me know. I decided to use local because Rufus is ultimately just using `Time.now`. I'm pretty sure it doesn't really matter as long as its consistent.

- Use `Fugit::Cron#match?` to check for exact time match. 
- Use `EtOrbi::EoTime#to_t` instead of `EtOrbi::EoTime#utc` to keep everything in local time.
- Use `Time#floor` instead of `Time#round`.
- Added specs for aperiodic jobs with and without time zones.

### DIFF
--- a/lib/sidekiq-scheduler/scheduler.rb
+++ b/lib/sidekiq-scheduler/scheduler.rb
@@ -267,10 +267,8 @@ module SidekiqScheduler
         if job_enabled?(name)
           conf = SidekiqScheduler::Utils.sanitize_job_config(config)
 
-          # `SidekiqScheduler::Utils.calc_cron_run_time` only works with a job running periodically.
-          # Also, `SidekiqScheduler::Utils.calc_cron_run_time` is only required jobs are execute frequently. So only call it if really needed.
-          if job.is_a?(Rufus::Scheduler::CronJob) && job.cron_line.rough_frequency <= 60
-            idempotent_job_enqueue(name, SidekiqScheduler::Utils.calc_cron_run_time(job.cron_line, time.utc), conf)
+          if job.is_a?(Rufus::Scheduler::CronJob)
+            idempotent_job_enqueue(name, SidekiqScheduler::Utils.calc_cron_run_time(job.cron_line, time.to_t), conf)
           else
             idempotent_job_enqueue(name, time, conf)
           end
--- a/spec/sidekiq-scheduler/utils_spec.rb
+++ b/spec/sidekiq-scheduler/utils_spec.rb
@@ -388,7 +388,7 @@ describe SidekiqScheduler::Utils do
       let(:cron) { Fugit::Cron.do_parse('* * * * *') }
 
       context 'when run at exact time' do
-        let(:time) { Time.utc(2024, 1, 1, 9, 0, 0) }
+        let(:time) { Time.new(2024, 1, 1, 9, 0, 0) }
 
         it 'returns the current time' do
           expect(described_class.calc_cron_run_time(cron, time)).to eq(time)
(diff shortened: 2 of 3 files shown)


marcelolx (approved):
This looks good to me. 

I will wait on @y-yagi and @muk-ai before merging and pushing a new release.

Thanks @afrase 🙏

lib/sidekiq-scheduler/scheduler.rb
```diff
@@ -267,10 +267,8 @@ def new_job(name, interval_type, config, schedule, options)
         if job_enabled?(name)
           conf = SidekiqScheduler::Utils.sanitize_job_config(config)
 
-          # `SidekiqScheduler::Utils.calc_cron_run_time` only works with a job running periodically.
-          # Also, `SidekiqScheduler::Utils.calc_cron_run_time` is only required jobs are execute frequently. So only call it if really needed.
-          if job.is_a?(Rufus::Scheduler::CronJob) && job.cron_line.rough_frequency <= 60
-            idempotent_job_enqueue(name, SidekiqScheduler::Utils.calc_cron_run_time(job.cron_line, time.utc), conf)
+          if job.is_a?(Rufus::Scheduler::CronJob)
+            idempotent_job_enqueue(name, SidekiqScheduler::Utils.calc_cron_run_time(job.cron_line, time.to_t), conf)
```
y-yagi, line 271: In my personal opinion, passing an instance of `EtOrbi::EoTime` as is better.

* `EtOrbi::EoTime` only holds a seconds. So no need to call `round`.
* `EtOrbi::EoTime` has `#-` and it allow to specify an instance of `EtOrbi::EoTime` . So we can get `next_t` and previous_t` without converting to `Time` instance. This allows us to reduce objects.
  * https://github.com/floraison/et-orbi/blob/ac15d51140720db6b82fa8e84b10d2f36f41ec24/lib/et-orbi/time.rb#L252

y-yagi (approved):
LGTM! Thanks!

spec/sidekiq-scheduler/utils_spec.rb
```diff
@@ -442,36 +442,129 @@
...
       end
     end
+
+    context 'when running aperiodically' do
+      context 'when a time zone is set' do
+        let(:cron) { Fugit::Cron.do_parse('0 8-18 * * * Asia/Tokyo') }
+
+        context 'when run at exact time' do
+          # Tokyo is +9 hours from UTC
+          let(:time) { Time.new(2024, 1, 1, 8, 0, 0, in: '+09:00') }
```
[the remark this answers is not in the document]
afrase: Thanks for pointing that out, I'll avoid the `in` param.

lib/sidekiq-scheduler/scheduler.rb
afrase: So `EtOrbi::EoTime` holds the seconds as a float and Rufus is using `EoTime.now` which uses `Time.now.to_f` which contains sub-seconds. Removing the sub-seconds was something I discovered after seeing multiple jobs get scheduled twice consistently 1 second after the scheduled time. So I think the removal of sub-seconds is important.

I do think passing `EtOrbi::EoTime` is better but the codebase is marked as expecting a `Time` object everywhere and uses `Time.now` as defaults. I think maybe just converting everything to `Time` in the gem is better to stay consistent. As far as I can tell `EtOrbi::EoTime` is only in this codebase in `SidekiqScheduler::Scheduler#new_job` block and `SidekiqScheduler::Utils.new_rufus_scheduler` in the `on_post_trigger` callback.

muk-ai:
LGTM👍
Thanks for taking the time to do this PR.

marcelolx:
Thanks to everyone! Released [v5.0.6](https://github.com/sidekiq-scheduler/sidekiq-scheduler/releases/tag/v5.0.6).

PR #490, opened 2024-10-02
updated example in README to use Sidekiq::Job

jcksnparsons:
This updates the documentation to use `Sidekiq::Job` instead of `Sidekiq::Worker`.

https://github.com/sidekiq-scheduler/sidekiq-scheduler/issues/489


.idea/workspace.xml
```diff
@@ -0,0 +1,66 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
marcelolx, line 1: Can we remove this file? ty!

jcksnparsons: Oops! Yep, my apologies!

marcelolx:
Thanks!

PR #491, opened 2024-11-21
Remove `tilt` from the dependency

y-yagi:
It seems that `tilt` is for `sinatra`.
https://github.com/sidekiq-scheduler/sidekiq-scheduler/commit/0313e5b#r14982005

I assume this was needed when the Sidekiq UI depended on `sinatra`, but that has not depended on it since
https://github.com/sidekiq/sidekiq/pull/3075.

So I think we don't need it now. Let's remove it.

### DIFF
--- a/lib/sidekiq-scheduler.rb
+++ b/lib/sidekiq-scheduler.rb
@@ -1,5 +1,4 @@
 require 'sidekiq'
-require 'tilt/erb'
 
 require_relative 'sidekiq/scheduler'
 require_relative 'sidekiq-scheduler/version'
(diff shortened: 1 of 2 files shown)


y-yagi:
Thanks for your review!

I'm sorry if I'm wrong, but Sidekiq removed Sinatra dependency at v4.2.0. 
https://github.com/sidekiq/sidekiq/blob/main/Changes.md#420

So I think this PR doesn't force users Sidekiq v7.

marcelolx:
oh nvm, I think I was looking at the wrong commit — perfect, this is even better!

PR #492, opened 2024-12-05
Adds visualization of disabled recurring jobs to Dark Mode

hmasing:
The current dark mode does not visually differentiate between enabled and disabled jobs.  The light mode version tints disabled jobs slightly differently.  

This patch adds this in to dark mode, which signficantly reduces the cognitive load when determining which jobs are active or disabled.

Super-simple patch, simply changes the text color of disabled jobs to a light gray instead of white.

sidekiq-scheduler-dark-mode-update


marcelolx:
Thanks @hmasing!

PR #495, opened 2025-03-07
Clean up checks and logic for older Sidekiq versions

t27duck:
Figured we may as well get the ball rolling.

This removes checks for older Sidekiq versions leaving support for only Sidekiq 7.3 which still has a level of support.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
       matrix:
         redis-version: ["~> 4.2", "~> 5"]
         ruby-version: ['2.7', '3.0', '3.1', '3.2', '3.3', '3.4']
-        sidekiq-version: ['~> 6', '~> 7']
+        sidekiq-version: ['~> 7.3']
     services:
       redis:
         image: redis
--- a/sidekiq-scheduler.gemspec
+++ b/sidekiq-scheduler.gemspec
@@ -18,7 +18,7 @@ Gem::Specification.new do |s|
 
   s.metadata = { 'rubygems_mfa_required' => 'true' }
 
-  s.add_dependency 'sidekiq', '>= 6', '< 8'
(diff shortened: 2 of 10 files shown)


sidekiq-scheduler.gemspec
```diff
@@ -30,4 +30,5 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'activejob'
   s.add_development_dependency 'rack-test'
   s.add_development_dependency 'rack', '< 3'
+  s.add_development_dependency 'rack-session'
```
t27duck, line 33: Sidekiq web in 8 requires Rack 3 which is its own can of worms. The session stuff was extracted into a separate gem so I figure may as well include it now.

.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       matrix:
         redis-version: ["~> 4.2", "~> 5"]
         ruby-version: ['2.7', '3.0', '3.1', '3.2', '3.3', '3.4']
-        sidekiq-version: ['~> 6', '~> 7']
+        sidekiq-version: ['~> 7.3']
```
t27duck, line 15: Sidekiq 8 only supports Ruby 3.2 and later, so this will be added probably once support for 8 is truly started:

```yaml
        exclude:
          - sidekiq-version: '~> 8'
            ruby-version: '2.7'
          - sidekiq-version: '~> 8'
            ruby-version: '3.0'
          - sidekiq-version: '~> 8'
            ruby-version: '3.1'
```

marcelolx:
Thanks for the PR @t27duck! I won't have time to review it for the next two weeks... but once I am back, I will take a look (quickly scanned through and it looks good)

marcelolx:
Also, feel free to move forward and merge it if you feel confident, so you don't have to wait on my review (once I am back to review it, if there is something missing we can adjust/fix as needed)

PR #497, opened 2025-03-13
Sidekiq 8 Support

denisahearn:
The changes in this PR make this gem compatible with Sidekiq 8

### DIFF
--- a/lib/sidekiq-scheduler/extensions/web.rb
+++ b/lib/sidekiq-scheduler/extensions/web.rb
@@ -2,21 +2,23 @@ require 'sidekiq/web' unless defined?(Sidekiq::Web)
 
 if SidekiqScheduler::SidekiqAdapter::SIDEKIQ_GTE_7_3_0
 
-  # Locale and asset cache is configured in `.register`
-  Sidekiq::Web.register(SidekiqScheduler::Web,
-    name: "recurring_jobs",
-    tab: ["Recurring Jobs"],
-    index: ["recurring-jobs"],
-    root_dir: File.expand_path("../../../web", File.dirname(__FILE__)),
-    asset_paths: ["stylesheets-scheduler"]) do |app|
-    # add middleware or additional settings here
+  # Locale and asset cache is configured in `cfg.register`
+  Sidekiq::Web.configure do |cfg|
+    cfg.register(SidekiqScheduler::Web,
+      name: "recurring_jobs",
+      tab: ["Recurring Jobs"],
+      index: ["recurring-jobs"],
+      root_dir: File.expand_path("../../../web", File.dirname(__FILE__)),
+      asset_paths: ["stylesheets-scheduler"]) do |app|
+      # add middleware or additional settings here
+    end
   end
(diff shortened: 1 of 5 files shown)


lib/sidekiq-scheduler/extensions/web.rb
```diff
@@ -1,11 +1,13 @@
...
+# Locale and asset cache is configured in `cfg.register`
+Sidekiq::Web.configure do |cfg|
+  cfg.register(SidekiqScheduler::Web,
+    name: "recurring_jobs",
+    tab: ["Recurring Jobs"],
+    index: ["recurring-jobs"],
+    root_dir: File.expand_path("../../../web", File.dirname(__FILE__)),
+    asset_paths: ["stylesheets-scheduler"]) do |app|
+    # add middleware or additional settings here
+  end
```
denisahearn, line 12: Clean up a `Sidekiq::Web.register is deprecated, use Sidekiq::Web.configure {|cfg| cfg.register(...) }` warning that Sidekiq emitted

lib/sidekiq-scheduler/web.rb
denisahearn, line 1: Clean up `Direct access to Rack parameters is discouraged, use url_params or route_params` warnings emitted by  sidekiq.

This also fixed a bug where clicking the `Enable/Disable` button for a scheduled job wasn't working because `params[:name]` was returning `nil`.

denisahearn:
@marcelolx @t27duck New PR based on the changes in https://github.com/sidekiq-scheduler/sidekiq-scheduler/pull/495 along with my changes to make this gem compatible with Sidekiq 8.

lib/sidekiq-scheduler/version.rb
```diff
@@ -1,3 +1,3 @@
 module SidekiqScheduler
-  VERSION = "5.0.6"
+  VERSION = "5.0.7"
```
denisahearn, line 2: Please let me know if you would prefer a different version number, like 5.1, etc.

t27duck:
Thanks @denisahearn I'll enable CI and like I said I'll try to carve out some time this weekend to poke around at it.

Regarding the version, I think there was talk on bumping the major version since the gem will be droping Sidekiq 6 and older Ruby support, but that'll be handled once the gem is ready for release.

marcelolx:
> Regarding the version, I think there was talk on bumping the major version since the gem will be droping Sidekiq 6 and older Ruby support, but that'll be handled once the gem is ready for release.

Correct, we already have a [v5-stable](https://github.com/sidekiq-scheduler/sidekiq-scheduler/tree/v5-stable) branch, so anything going to master is for the new release and will result in v6 of this gem

t27duck:
Confirmed stuff at least functions on Sidekiq 8 with this change along with specs. Specs and support for Sidekiq 7.3 is a little on the broken side, though. Going to (soon) merge this as-is and apply a patch separately to get CI to pass and bring back 7.3 support.

Appreciate the work, @denisahearn

PR #520, opened 2026-08-06
Batch recurring job metadata reads

OskarEichler:
## Summary

- batch recurring-job state, last-run, and next-run reads into one Redis pipeline
- pass the preloaded metadata into each `JobPresenter` while preserving the existing direct-instantiation fallback
- evaluate `enabled?` once per row in both supported Sidekiq Web templates

## Why

The recurring jobs page currently reads Redis separately every time a presenter renders its enabled state, last execution time, or next execution time. The Sidekiq 8 template calls `enabled?` three times per job, so a page with 236 recurring jobs performs about 1,180 presenter-level Redis reads.

This change replaces those per-job reads with three `HMGET` commands sent through one pipeline. Directly instantiated presenters retain the existing behavior and continue loading their own values from Redis.

## Validation

- Sidekiq 8 / Rack 3 full suite: 277 examples, 0 failures
- Sidekiq 7.3 / Rack 2.2 focused Redis manager, presenter, and web specs: 68 examples, 0 failures
- Ruby syntax and `git diff --check`

### DIFF
--- a/lib/sidekiq-scheduler/job_presenter.rb
+++ b/lib/sidekiq-scheduler/job_presenter.rb
@@ -11,16 +11,19 @@ module SidekiqScheduler
 
     include Sidekiq::WebHelpers
 
-    def initialize(name, attributes)
+    def initialize(name, attributes, metadata: nil)
       @name = name
       @attributes = attributes
+      @metadata = metadata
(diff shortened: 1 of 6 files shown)


OskarEichler:
@marcelolx

PR #525, opened 2026-08-27
Apply listened_queues_only to dynamic schedule updates

### LINKED ISSUE #521: Dynamic schedule updates ignore listened_queues_only
## Summary
With `listened_queues_only: true`, startup ignores jobs for queues this process does not listen to, but `update_schedule` subsequently loads those same jobs without applying the filter.

## Reproduction
Verified with sidekiq-scheduler 6.0.2, Sidekiq 8.1.7 and Ruby 4.0.6, using only disposable local Redis.
1. Configure the process queues as `['default']` and scheduler `listened_queues_only: true`.
2. Store a schedule named `audit-queue` with `class: 'AuditWorker', every: '1d', queue: 'other'`.
3. Call `load_schedule!`: `scheduled_jobs.keys` is empty, as expected.
4. Poll a change window containing that schedule through `update_schedule`.
5. `scheduled_jobs.keys` becomes `['audit-queue']`.

The initial load checks `enabled_queue?`; the dynamic update calls `load_schedule_job` directly. A job whose queue changes from an allowed queue to an excluded queue should also be unscheduled rather than reloaded. This reproduces on the installed release; these methods are unchanged on current master.

Expected: the configured queue filter applies to both initial and dynamic loading. No intended change to the existing empty-queues policy.

Found and reproduced with Codex assistance. A focused source fix is being prepared; no production diagnostics or operations were used.

OskarEichler:
## Summary
Use the same queue-filtering helper for initial loading and dynamic changes. Compute the configured queues once per batch, and keep the existing unschedule-before-reload behavior so moving a job to an excluded queue removes it.

Fixes #521. The issue includes reproduction steps and original behavior.

## Verification
- Ruby 4.0.6, Sidekiq 8.1.7, Rack 3.2.7 and Rufus 3.9.2.
- Initial excluded job stays excluded after polling; moving it to an allowed queue adds it; moving it back removes it. Empty configured queues still allow all jobs, and deletion removes the schedule. Existing disabled-filter behavior stays unchanged.
- Full existing upstream RSpec suite on the unchanged master baseline and this individual patch: **277 examples, 0 failures**.
- Reproductions use bounded future schedules and shut down their Rufus instances; Redis is a disposable local service over a private Unix socket.
- Ruby syntax and whitespace checks pass. Targeted Lint checks use a temporary external configuration, excluding two pre-existing offenses in unchanged code (NonLocalExitFromIterator and UselessConstantScoping). No upstream lint configuration changed.

## Compatibility and limitations
No intended breaking changes. No runtime dependency or version changes. Other Sidekiq/Ruby/platform combinations and upstream CI are not yet verified locally for this individual patch. No test files added or modified, per the originating project's policy; focused repros ran outside the repository.

Prepared with Codex assistance. Findings were reproduced locally; no production access or deployment was used.

### DIFF
--- a/lib/sidekiq-scheduler/scheduler.rb
+++ b/lib/sidekiq-scheduler/scheduler.rb
@@ -97,11 +97,7 @@ module SidekiqScheduler
         queues = scheduler_config.sidekiq_queues
 
         Sidekiq.schedule.each do |name, config|
-          if !listened_queues_only || enabled_queue?(config['queue'].to_s, queues)
-            load_schedule_job(name, config)
-          else
-            Sidekiq.logger.info { "Ignoring #{name}, job's queue is not enabled." }
-          end
+          load_schedule_job_for_queues(name, config, queues)
         end
 
         Sidekiq.logger.info 'Schedules Loaded'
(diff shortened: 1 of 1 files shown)


marcelolx:
@OskarEichler I will see if I can review those PRs this week. One question, have you been actively using sidekiq-scheduler and running it with those fixes?

